PR #91, opened 2022-01-29
Power rework

GermanHydrogen:
## Framework
- [x] Compiler
- [x] Device
- [ ] Battery
- [x] Generator
- [ ] Consumer

## Functionality
- [ ] Power Capacity

## Objects
- [ ] Generator (Update)
- [ ] Laptop (Update)
- [ ] Battery
- [ ] Lamp (?)


y0014984 (commented):
I have reviewed your Pull Request. I don't understand everything you made but I think I have a basic understanding of the power rework Pull Request. I have some annotations:

- please use Land_BatteryPack_01_open_sand_F instead of Land_BatteryPack_01_open_black_F because I used Sand as the default color. We will later add color variations.
- The battery can be turned on, even it is completely empty. This should not be possible.
- Battery and Laptop initialize with emty batteries. They should be fully loaded like the generator.
- I can connect the empty laptop to the full battery but the laptop does not load. Booting/starting is also not possible.
- We should rename "Connect to Generator" to "Connect to Power Source", if generator and battery can be power sources.
- The Eden Editor AE3 Attributes "Power LEvel" and "Fuel Level" doesn't work anymore. Perhaps they are broken for a long time.
- ae3_power_batteryCapacity and ae3_power_recharging are replaced by capacity and recharging in AE3_Battery. The values doesn't match, see https://docs.google.com/spreadsheets/d/1uKQK8no6wsR7n2-Xa8Vi-e6YC-xwoeOwcBT4-IKHopQ/edit#gid=0
- Charging speed is too fast. A charging value 0.1 means with a 1sec eachFrameEventHandler, the device is fully loaded in 100 sec. If laptop loads with 50 Watts and has 100 Wh capacity, then I would say, loading in device off state should take 2 hours, leading to a charging value of approx. 0.0139
- Laptop has two different ACE interaction menus called "Laptop". One for using laptop and the other one for managing power. We should combine them.
- The battery capacity in my former mentioned list is always in Wh, so there is no need to multiply by 1000 in the hint. You only need to increase the capacity value from 10 to 100

Issue #86, opened 2022-01-08
ArmaOS User folder creation


GermanHydrogen:
Create the user folder in home/ when a new user is created.

Issue #85, opened 2022-01-08
ArmaOS Function Rework


GermanHydrogen:
The current system is not very flexible and adding a new function requieres editing unrelated code in many places.

Changes:

- [x] Abstracting functions to executable files with global links
- [x] Defining global links in seperate config

y0014984:
I'm not sure what this ticket is about. Was this part of your already done rework or is this something new, like custom functions that can be executed in ArmaOS directly and having a filesystem representation?

GermanHydrogen:
There is no real flexible function system implace. At the momentent all ArmaOS functions are hardcoded.

Issue #84, opened 2022-01-08
ArmaOS Filesystem Permission System


GermanHydrogen:
- [ ] File Ownership
- [ ] Permission System

Issue #60, opened 2020-12-21
Change default Module Attributes


y0014984:
5 filled in attributes are to much. I think that it's enough if the first attribute is filled in. I need to check, if the attributes are blank and don't process them.

PR #94, opened 2022-02-21
Armaos functions and permission

GermanHydrogen:
Added function abstraction #85 
Added file permission system #84
Added User folder creation #86 
Fixed Eden Module Attributes #60


addons/armaos/functions/fnc_shell_executeFile.sqf
```diff
@@ -0,0 +1,42 @@
...
+{
+	_content = [_pointer, _filesystem, _path, _username, 0] call AE3_filesystem_fnc_getFile;
+
+	if(_content isEqualType {}) then
+	{
+		_result = [_computer, _options] call _content;
+
+		if(isNil "_result") exitWith {_result = [""]};
+
+		if(_result isEqualType []) exitWith {};
```
y0014984, line 34: I think we need a result in this case, like in the other cases, for example:

`if(_result isEqualType []) exitWith {_result = [""]};`

GermanHydrogen: exitWith in this case only breaks out of the try catch context.
The result is set in line 30 and does not to be modified if its a list.

y0014984: Ok. I see. By using `exitWith` you only leave the `try` scope. So `_result` is already set and will be returned in lin 42.

y0014984 (commented):
I have done a code review. No I will test the modifications in-game.

y0014984:
Please rename user module: AE3 Userlist --> AE3 Add User
After that you can merge the pull request. 

Some additional annotations for further enhancements:

If we have a permission system, we need some additional commands and/or information provided to the user:
- List privileges and owner with ls (Example: `[root rwx---] password.txt`)
- Change owner with `chown`
- Change privileges with `chmod`
- Check active user with `whoami` or show in prompt (Example: `user@host:/path/to/>`)
- Deny reading/accessing folders with `ls/cd` with insufficient read permissions
- List available users with `users` or `cat etc/passwd` or `getent passwd` or `ls /users/`
- Create home directories and users for files without corresponding users (so `ls /users/` would work)

AE3 Add File Module:
- is code? is somewhat redundant to the execute privilege (In case someone creates a text file and changes the permissions to be executeable, it shouldn't be necessary to set this additional flag)

Additional and useful commands:
- Change password with passwd
- Get root privileges with sudo (checkbox in Add User Module for sudoer list)
- Change user context with su

PR #99, opened 2022-02-24
Structured text terminal

GermanHydrogen:
Changed main terminal dialog textfield to structured text.


y0014984 (approved):
I have no additional annotations. You can merge this PR. 

Later we need to review the Terminal Buffer. Lines that are added to the buffer are cropped, resulting in additional lines. This cropping is done on the basis of a max. column value. If the text contains additional informations for structured text, these lines will be cropped to early. But for now we can leave this as it is.

PR #100, opened 2022-03-05
ArmaOS whoami

GermanHydrogen:
- Added whoami command
- Added user and hostname to terminal prompt (#63)


y0014984 (approved):
You can merge this PR. Funny mechanism to use the ACE Rename Feature for hostname. As a later feature we could allow the mission creator to set the hostname, perhaps as an attribute in the laptop asset. At this stage I would also prefer to randomize the hostnames, if not set in init phase, so every computer has a unique hostname. This will help us with networking. But for now this is enough.

PR #101, opened 2022-03-05
Added long mode to ls

GermanHydrogen:


y0014984 (approved):
I have no annotations. You can merge this PR. An additional sorting and aligning algorythm at the end of processing would be nice, so the output is sorted alphabetically and usernames and filenames are aligned, but this a later feature and not mandatory for this PR.

y0014984 (approved):
Sorting and alignment looks great. You can merge this PR.

Issue #145, opened 2022-04-02
Getter for the direction of solar panels


y0014984:
@GermanHydrogen needs a getter function or accessable variable to know what the absolute direction of a specific solar panel is. This is needed for the calculation of the resulting power output according to sun position.

PR #147, opened 2022-04-19
Solarpanels

GermanHydrogen:
Added power generation to rugged and flexible solar panels.

Closing #145 
Closing #3

### DIFF
--- a/addons/power/CfgVehicles.hpp
+++ b/addons/power/CfgVehicles.hpp
@@ -216,7 +216,7 @@ class CfgVehicles
 			{
 				powerMax = 0.1/3600; // In this case per panel
 				orientationFnc = "_this call AE3_power_fnc_multSolarPanelOrientation";
-				height = 1.5;
+				height = 1.2;
 			};
 		};
 
--- a/addons/power/CfgVehicles.hpp
+++ b/addons/power/CfgVehicles.hpp
@@ -219,6 +219,28 @@ class CfgVehicles
 				height = 1.5;
 			};
 		};
+
+		class AE3_InternalDevice
+		{
+			displayName = "Battery";
+			defaultPowerLevel = 1;
+
+			turnOnAction = "_this + [true] call AE3_power_fnc_turnOnBatteryAction";
+			turnOffAction = "";
+
+			class AE3_PowerInterface
+			{
+				internal = 0;
+			};
+
+			class AE3_Battery
+			{
+				capacity = 0.4;
+				recharging = 0.05/3600;
+				level = 0;
+				internal = 1;
+			};
+		};
 	};
 
 
--- a/addons/power/CfgVehicles.hpp
+++ b/addons/power/CfgVehicles.hpp
@@ -135,7 +135,7 @@ class CfgVehicles
 
 	/* ================================================================================ */
 
-	// Solar Panel
+	// Solar Panels
 	class Land_SolarPanel_04_sand_F;
 
 	class Land_SolarPanel_04_sand_F_AE3 : Land_SolarPanel_04_sand_F
@@ -166,9 +166,9 @@ class CfgVehicles
 					{
 						description = "pitch solar panel 1";
 						animation = "Panel_1_Pitch";
-						minValue = -90;
-						maxValue = 90;
-						scrollMultiplier = 10;
+						minValue = -45;
+						maxValue = 45;
+						scrollMultiplier = 5;
 					};
 				};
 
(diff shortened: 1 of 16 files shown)


GermanHydrogen:
@y0014984 Is it possible to move the ace interaction points for the animations?
The interaction point of the yaw axis of the rugged solar panel is very close to the objects main interaction point.

y0014984:
> @y0014984 Is it possible to move the ace interaction points for the animations? The interaction point of the yaw axis of the rugged solar panel is very close to the objects main interaction point.

We are using Selection Names for dynamic interaction positions. That makes sense for the rugged lights, because these selection positions use dynamically the actual hight of a lamp. But for other equipment this doesn't fit very well. Therefore we have two options:

Option A: Use the Main Interaction Point
Option B: Define a custom iteraction position that is different from the main interaction point

@GermanHydrogen please choose your prefered option and assign me a ticket. I will make the adjustment as soon as I can.

y0014984:
I will do the review on friday.

y0014984:
@GermanHydrogen : I tested the solar panel output, especially the modification by visibility of the sun. In full shadow behind a building the panel produced full power output. So I checked the code myself and it seems that the virtual position of the sun is horizontally approx. 90 degrees off it's current position. I tried that on altis and tanoa in March and November. It's mostly 90 degrees offset to the east, see screenshots. In October on Altis the Offset was more something like 150 degrees.

https://i.imgur.com/0P9AH1P.jpg
https://i.imgur.com/Nl6PFHp.jpg

Additionally you can see that the calculated distance for the virtual sun is only about 5-10 meters away from the solar panel. A higher Value would be better in the vectorMultiply function so the virtual sun is something like 50 to 100 meters away. You use the unmodified position of the solar panel to calculate the position of the virtual sun, but you shoud use the modified (lifted) position for exact results, see screenshot.

You could interchange lines 29 and 30 and modifiy the new line 30 to look like

```sqf
private _solPos = _panelPos vectorAdd (_solVec vectorMultiply 10);
```

addons/power/CfgVehicles.hpp
```diff
@@ -196,13 +196,82 @@ class CfgVehicles
...
+			defaultPowerLevel = 0;
+
+			turnOnAction = "_this call AE3_power_fnc_turnOnSolarAction";
+			turnOffAction = "_this call AE3_power_fnc_turnOffSolarAction";
+
+			class AE3_SolarGenerator
+			{
+				powerMax = 0.1/3600; // In this case per panel
+				orientationFnc = "_this call AE3_power_fnc_multSolarPanelOrientation";
+				height = 1.5;
```
y0014984, line 219: 1.2 fits better

addons/power/functions/fnc_solarCalculation.sqf
```diff
@@ -0,0 +1,35 @@
...
+{
+	_buffer = _x vectorCos _solVec;
+	if (_buffer > 0) then 
+	{
+		_mult = _mult + _buffer;
+	};
+}forEach _normalList;
+
+// If sun is visible
+private _solPos = getPosASL _solar vectorAdd (_solVec vectorMultiply 10);
```
y0014984, line 29: //Move down one line and modify likes this:
private _solPos = _panelPos vectorAdd (_solVec vectorMultiply 10);

y0014984: vectorMultiplay 100 would be better

addons/power/functions/fnc_turnOffSolarAction.sqf
```diff
@@ -0,0 +1,35 @@
...
+		{
+
+		},
+		{},
+		("Turn Off")
+	] call ace_common_fnc_progressBar;
+};
+
+[_entity, true, [0, 1, 0], 0] call ace_dragging_fnc_setDraggable;
+
```
y0014984, line 27: //We should add something like this, so the checkPowerGenAction will return 0 Watts if the solar panel is powered down.

_entity setVariable ['AE3_power_powerCapacity', 0, true];

y0014984 (commented):
Besides my comments I have no additional things to say. If these things get fixed then you can merge this PR.

y0014984 (commented):
I have tested the sun calculations on 17 maps, official and custom ones. (Altis, Stratis, Malden, Livonia, Tanoa, Virtual Reality, Cam Lao Nam, Khe San, Weferlingen, Sefron Ramal, Chernarus, Kujari, Lythium, Rosche, Ruha, Takistan, Virolahti). On all maps the sun calculations worked very well. Sometimes the virtual sun seems to be a little bit too high but thats not a big deal. Only on tanoa the offsets were bigger. Still usable but i qould say a few degrees offset. Perhaps this is due to the position of tanoa, that is very close to the equator. To sum it all together I would say we have very good sun calculations to work with.

The power production on the other hand is a problem as long as we still use the sunOrMoon command to calculate the power output instead of the sun angle and visablitity of the sun. The solar panel still produces power (loads it's internal battery) in very dark (not night) conditions or in full shadow behind buildings. If you check Power Output, it says 0 Watts, but the battery is still loading. @GermanHydrogen is this the intended behavior?

y0014984:
Looks good now. I will merge.

Issue #142, opened 2022-04-02
Closed terminal on shutdown or standby


y0014984:
Terminal is closed on shutdown or standby due to some ACE progressBar magic. To be sure, the terminal window should be closed by script, not by accident.

Issue #117, opened 2022-04-02
In multiplayer environment laptop is useable simoultaniously by all players


y0014984:
We should restrict the usage of the laptop/armaOS to one player at a time. This could be achieved by tracking the active user/player. If a player is using the laptop, no one else should be able to access it.

Issue #96, opened 2022-02-22
Reset currend working directory on user switch


y0014984:
If you logout from ArmaOS terminal and login to another user, the working directory is still the same. This is a security issue.

Issue #95, opened 2022-02-22
Disable Keyboard Layout Auto-Switching


y0014984:
After leaving and using Terminal in ArmaOS again, the keyboard layout switches automatically. This is a bug. The current keyboard layout needs to be saved and loaded.

GermanHydrogen:
This is now fixed in #151

PR #151, opened 2022-10-05
Various armaos bug fixes

GermanHydrogen:
Closing #142 
Closing #117 
Closing #96 
Closing #95 
Closing #159 
Closing #160 
Closing #158

### DIFF
--- a/addons/armaos/CfgVehicles.hpp
+++ b/addons/armaos/CfgVehicles.hpp
@@ -109,7 +109,9 @@ class CfgVehicles
 			init = "_this call AE3_interaction_fnc_initLaptop;";
 
 			openAction = "_this call AE3_interaction_fnc_laptop_open;";
+			openActionCondition = "isNull (_this getVariable ['AE3_computer_mutex', objNull])";
 			closeAction = "_this call AE3_interaction_fnc_laptop_close;";
+			closeActionCondition = "isNull (_this getVariable ['AE3_computer_mutex', objNull])";
 		};
 
 		class AE3_Device
@@ -120,8 +122,11 @@ class CfgVehicles
 			init = "_this call AE3_filesystem_fnc_initFilesystem; _this call AE3_armaos_fnc_link_init; _this call AE3_network_fnc_initNetworkDevice;";
 
 			turnOnAction = "_this call AE3_network_fnc_dhcp_onTurnOn; _this call AE3_armaos_fnc_computer_turnOn;";
+			turnOnActionCondition = "isNull (_this getVariable ['AE3_computer_mutex', objNull])";
 			turnOffAction = "_this call AE3_armaos_fnc_computer_turnOff;";
-			standByAction = "_this call AE3_armaos_fnc_computer_standby;";
+			turnOffActionCondition = "isNull (_this getVariable ['AE3_computer_mutex', objNull])";
+			standByAction = "isNull (_this getVariable ['AE3_computer_mutex', objNull])";
+			standByActionCondition = "false";
 
 			class AE3_Consumer
 			{
--- a/addons/armaos/CfgVehicles.hpp
+++ b/addons/armaos/CfgVehicles.hpp
@@ -167,8 +167,8 @@ class CfgVehicles
 					class AE3_UseComputer
 					{
 						displayName = "Use";
-						condition = "(alive _target) and (_target getVariable 'AE3_power_powerState' == 1)";
-						statement = "params ['_target', '_player', '_params']; _handle = [_target] call AE3_armaos_fnc_terminal_init;";
+						condition = "(alive _target) && (_target getVariable 'AE3_power_powerState' == 1) && (isNull (_target getVariable ['AE3_computer_mutex', objNull]))";
+						statement = "params ['_target', '_player', '_params']; _target setVariable ['AE3_computer_mutex', _player, true]; _handle = [_target] call AE3_armaos_fnc_terminal_init;";
 						//icon = "\z\dance.paa";
 						exceptions[] = {};
 						//insertChildren
--- a/addons/armaos/functions/fnc_computer_standby.sqf
+++ b/addons/armaos/functions/fnc_computer_standby.sqf
@@ -12,8 +12,7 @@ params ["_computer"];
 
 [_computer] spawn {
 	params ["_computer"];
-
-	sleep 3;
+	
 	_computer setObjectTextureGlobal [1, "\z\ae3\addons\armaos\textures\Laptop_4_to_3_Standby.paa"];
 
 	[_computer] call AE3_armaos_fnc_computer_playSoundStandby;
--- a/addons/armaos/functions/fnc_computer_standby.sqf
+++ b/addons/armaos/functions/fnc_computer_standby.sqf
@@ -18,5 +18,5 @@ params ["_computer"];
 
 	[_computer] call AE3_armaos_fnc_computer_playSoundStandby;
 };
-
+closeDialog 2;
 true;
--- a/addons/armaos/functions/fnc_computer_standby.sqf
+++ b/addons/armaos/functions/fnc_computer_standby.sqf
@@ -10,23 +10,13 @@
 
 params ["_computer"];
 
-private _standbyTime = 3;
+[_computer] spawn {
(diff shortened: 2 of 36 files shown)


addons/armaos/CfgVehicles.hpp
```diff
@@ -167,8 +167,8 @@ class CfgVehicles
 					class AE3_UseComputer
 					{
 						displayName = "Use";
-						condition = "(alive _target) and (_target getVariable 'AE3_power_powerState' == 1)";
-						statement = "params ['_target', '_player', '_params']; _handle = [_target] call AE3_armaos_fnc_terminal_init;";
+						condition = "(alive _target) && (_target getVariable 'AE3_power_powerState' == 1) && (isNull (_target getVariable ['AE3_computer_mutex', objNull]))";
```
y0014984, line 170: If this is kind of a locking mechanism, I think it would be better to additionally check if player == AE3_computer_mutex, especially in cases where the player stops using armaOS not by triggering shutdown or standby, for example a disconnect or some weird interface thingy that allows you to close the dialog in a bypass kind of way. In this case the player could return to his own "session". What do you think?

addons/armaos/functions/fnc_computer_standby.sqf
```diff
@@ -10,23 +10,14 @@
...
+	params ["_computer"];
 
-[
-	_standbyTime,
-	[_computer], 
-	{
-		params ["_args", "_elapsedTime", "_totalTime", "_errorCode"];
-		
-		private _computer = _args select 0;
+	sleep 3;
```
y0014984, line 16: If you disable the progessBar for the standby action, why using sleep command? If there is no progess, the sound could play immediately.

addons/armaos/CfgVehicles.hpp
GermanHydrogen: The lock is removed, if the dialog is closed. This should in theory also cover cases, when the dialog is
closed unintentionally. 
Because the player object is referenced in the variable, it should also become null, if the player
disconnects from the server. So the lock is also lifted in this case.

addons/armaos/functions/fnc_computer_standby.sqf
GermanHydrogen: fixed

addons/armaos/CfgVehicles.hpp
GermanHydrogen: https://github.com/y0014984/Advanced-Equipment/blob/674583d280f7f4a7310168659b1a17d0b9445edd/addons/armaos/functions/fnc_terminal_addEventHandler.sqf#L204

addons/armaos/functions/fnc_computer_standby.sqf
y0014984: Ok. I think I understand now. You removed the progess bars completely. I presume you done that due to immersion reasons because the display on the laptop already gives a visual indicator. So the time is the previous time of the progress bar.

GermanHydrogen: I thought that we agreed on that in the past. The arguments for the removal were, that its quiete annoying to wait and watch the progress bar for 15s  and that in rl, you also dont have to hug your computer for the time it is booting.

addons/armaos/CfgVehicles.hpp
y0014984: I saw you already managed this event by adding the Display Unload Event Handler. I think this would be sufficient.

y0014984 (commented):
I went through all the changes. Nothing to mention at the moment. I will test the changes now in dedicated multiplayer.

y0014984:
Findings in dedicated multiplayer testing environment:

- While the laptop is carried by player A, player B can access and execute these ACE interactions: open, close, check power state, check battery level, turn on/off; Acessing ArmaOS is not possible; Perhaps we need an additional mutex for carrying the laptop
- Turn on/off and standby is now disabled while the laptop is in use
- Carrying the laptop still works for player B, while player A is using the laptop (uses ArmaOS interface); Checking power state and battery charge is also still available, but turn off and standby are disabled
- The laptop now remembers the last used keyboard layout, even if the laptop is turned off and an again; The setting seems to be user depended; If player A changes from default US to DE on a running laptop, then for player B the US keyboard layout is still present; It's a question of game design and simulation, if the setting should be player- oder device-related
- The standby action is not available anymore; I think it's due to lines 128 ans 129 in CfgVehicles.hpp in armaos; These lines do not follow the same syntax as for turn on or turn off
- If the laptop is closed while armaOS is still running, you get an script error:
14:47:04 Error in expression <isNull (_this getVariable ['AE3_computer_mutex', objNu>
14:47:04   Error position: <getVariable ['AE3_computer_mutex', objNu>
14:47:04   Error getvariable: Type Array, expected Namespace,Object,Group,Display (dialog),Control,Team member,Task,Location
- Opening up the laptop again shows running armaOS but I think it should be on standby
- File Permissions for files created by module seem to be fixed
- User folder seem to work
- Typing "standby" in armaOS also produces script error:
14:54:01 Error in expression <isNull (_this getVariable ['AE3_computer_mutex', objNu>
14:54:01   Error position: <getVariable ['AE3_computer_mutex', objNu>
14:54:01   Error getvariable: Type Array, expected Namespace,Object,Group,Display (dialog),Control,Team member,Task,Location
- RPT-File shows a lot of Performance warnings like these:
14:38:20 Performance warning: SimpleSerialization::Read 'ae3_terminal' is using type of ,'SCRIPT' which is not optimized by simple serialization, falling back to generic serialization, use generic type or ask for optimizations for these types
14:38:20 Performance warning: SimpleSerialization::Read 'ae3_terminal' is using type of ,'CONTROL' which is not optimized by simple serialization, falling back to generic serialization, use generic type or ask for optimizations for these types
14:54:01 Performance warning: SimpleSerialization::Write 'ae3_terminal' is using type of ',SCRIPT' which is not optimized by simple serialization, falling back to generic serialization, use generic type or ask for optimizations for these types
14:54:01 Performance warning: SimpleSerialization::Write 'ae3_terminal' is using type of ',CONTROL' which is not optimized by simple serialization, falling back to generic serialization, use generic type or ask for optimizations for these types

y0014984:
If have some questions:

- Why the booting sequence is now much faster than the shutdown sequence? You reduced booting from 15 to 3 sec.
- Why you are saving the variable "AE3_computer" to consoleOutput and to consoleDialog in terminal_init? If I remenber correctly you can access both if you only have one of them, for example:
`private _consoleOutput = _consoleDialog displayCtrl 1100;`
or
`_display = ctrlParent _control;`

y0014984:
I think that not all findings need to be fixed in this PR. I think the only relevant error is the missing/malfunctioning standby. If this is fixed then you can merge this PR.

GermanHydrogen:
> While the laptop is carried by player A, player B can access and execute these ACE interactions: open, close, check power state, check battery level, turn on/off; Acessing ArmaOS is not possible; Perhaps we need an additional mutex for carrying the laptop

Mutex on carrying is complicated, because the laptop just uses the ace native carrying action.

GermanHydrogen:
Could not reproduce:
> Typing "standby" in armaOS also produces script error:

GermanHydrogen:
> If the laptop is closed while armaOS is still running, you get an script error:

Does this mean, that another player was able to close the laptop, while another player used the terminal, or that the laptop was just on? @y0014984

y0014984:
> > If the laptop is closed while armaOS is still running, you get an script error:
> 
> Does this mean, that another player was able to close the laptop, while another player used the terminal, or that the laptop was just on? @y0014984

Can't reproduce anymore.

y0014984:
>

y0014984:
> Could not reproduce:
> 
> > Typing "standby" in armaOS also produces script error:

Me too.

y0014984:
> > Opening up the laptop again shows running armaOS but I think it should be on standby
> 
> Opening the laptop after standby and closing reboots the laptop automatically. Is this not the intendet behaviour? @y0014984

Yes, this is the intended behavior, but that wasn't the case. Perhaps my problem with standby was due to a build without deleting the previous build. I think we can ignore that. The standby-related issues seem to work now.

Issue #122, opened 2022-04-02
Generator sound bug


y0014984:
In dedicated multiplayer, if player1 turns on the generator and player2 turns off the generator, the sound is still playing for player1. @GermanHydrogen already mentioned that there is another way of sound playback for vehicles.

Issue #92, opened 2022-02-18
Disturbing Generator sound tabbing out and back into Arma 3


y0014984:
Perhaps this issue is related to the schedulder. @GermanHydrogen guesses, that a Run-On-Each-Frame Environment could help.

PR #161, opened 2022-10-11
Generator Fixes

GermanHydrogen:
Changed generator parent to B_Radar_System_01_F, which is better suited for a stationary object.
EngineOnSound is now handled by the engine itself.
Added dummy generator objects which appear in the correct editor category.

Closing #92 
Closing #122


y0014984 (changes requested):
Your technique of replacing objects is very nice. But there are still some issues we need to talk about:

- The class "B_Radar_System_01_F" is no vehicle and therefore has no fuel management. So fuel doesn't get lower and you can't refuel. I'm not sure if we can add this again on this class. Must be done in the config, I think. It's parents are static weapons. (["Radar_System_01_base_F","StaticMGWeapon","StaticWeapon","LandVehicle","Land","AllVehicles","All"])
- The Generator is now under Blufor vehicles and under Props. We need to check if the replacement class "B_Radar_System_01_F" could be hidden from the asset browser.
- One drawback changing to Props is, that the "Edit Vehicle Appearence..." Function can'T be used for changing texture. But that's a minor drawback if the benefits outweigh the disadvantages. 
- Running sound is much louder now than starting and stopping sound. Perhaps we can adjust this. 
- Do you start sounds for starting and stopping twice? I thought that the remoteCall of "eningeOn" would start the engine and the sound, but you also use playSound3D. Shouldn't be one of both solutions enough or does the engineOn command doesn't work? Could be because it isn't a vehicle with an engine. See fuel.

So what are the main benefits of changing the class?

GermanHydrogen:
Did you actually test the branch ingame?
The only valid problem is the sound, but fuel managment fully works.
The radar prop has the upside of having no inventory and no cargo spaces for players.
It still is a valid vehicle with engine and fuel simulation, but has no tiers which where one of the problems with the quad as a parent class.

The generator was under blufor vehicles before, when we the quad was the parent class.
This is mitigated by the indruduction by the dummy class which will appear under the right category for things.

The start and stop sound are still played via the script commands, because if they are specified in the config
with soundEngineOnExt and soundEngineOffExt, they will play simutaneously to the normal engine sound if the engine is started. The work around is, to play the start and stop sounds via script and then start the engine.
This is still better, then playing the engine sound via script, because this lead to various bugs, such as mentioned above.

PR #173, opened 2022-10-18
Latest-fixes-in-release-050

y0014984:


addons/armaos/functions/fnc_shell_findLoginUser.sqf
```diff
@@ -17,16 +17,25 @@ private _users = _computer getVariable ["AE3_Userlist", createHashMap];
 
 private _result = [];
 
-if (_username in _users) then 
+if (_username == "root") then
```
GermanHydrogen, line 20: Permits root access in general. 
To fix:
```
if (_username in _users) then
{...}else
{
   if (_username isEqualTo "root") then
   {...};
};
```

addons/armaos/XEH_preInit.sqf
```diff
@@ -1,2 +1,19 @@
...
+		["US", "DE"],
+		[["US", "United States"], ["DE", "Deutschland"]], 
+		0
+	],
+    nil, // "_isGlobal" flag. Set this to true to always have this setting synchronized between all clients in multiplayer
+    {  
+        params ["_value"];
+    }, // function that will be executed once on mission start and every time the setting is changed.
+    true // Setting will be marked as needing mission restart after being changed. (optional, default false) <BOOL>
+] call CBA_fnc_addSetting;
```
GermanHydrogen, line 19: This is set per mission/server right?

GermanHydrogen (changes requested):

addons/armaos/XEH_preInit.sqf
y0014984: All CBA settings are client-side by default as long as `_isglobal == false`, but mission makers and server admins can decide to overwrite client-settings with their mission or server settings, see [here](https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System).

addons/armaos/functions/fnc_shell_findLoginUser.sqf
y0014984: I will change that.

GermanHydrogen (approved):

PR #187, opened 2022-10-28
added/changed CBA-settings

y0014984:
- synced the keyboard layout in CBA settings to button in the terminal window
- added debug mode setting
   - info about activated debug mode in systemChat
   - add your own code to debug mode while loop
   - check in-game if debug mode is enabled via variable "AE3_DebugMode"
- added scroll speed setting


addons/armaos/functions/fnc_terminal_init.sqf
```diff
@@ -72,8 +72,20 @@ private _localGameLanguage = language;
 // if the language is german, it's obvious, that the keyboard layout is also german (this is not the case, if game language is english)
 // perhaps we need to provide a CBA setting for changing keyboard layout or allow to change the layout directly in the terminal window
 
+private _terminalKeyboardLayout = "";
+
 //AE3_KeyboardLayout is a CBA setting
-private _terminalKeyboardLayout = AE3_KeyboardLayout;
+if (!isMultiplayer || (isServer && hasInterface)) then
```
GermanHydrogen, line 78: `!isMultiplayer` should be redundand

addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
```diff
@@ -12,4 +12,16 @@ _terminal set ["AE3_terminalAllowedKeys", _terminalAllowedKeys];
 _languageButton ctrlSetText "KEYBOARD " + _terminalKeyboardLayout;
 ctrlSetFocus _consoleOutput;
 
+// write/sync changed keyboard layout back to CBA settings
+if (!isMultiplayer || (isServer && hasInterface)) then
```
GermanHydrogen, line 16: !isMultiplayer should be redundand

addons/main/functions/fnc_manageDebugMode.sqf
```diff
@@ -0,0 +1,23 @@
+params ["_debugMode"];
```
GermanHydrogen, line 1: Missing docstring

addons/main/functions/fnc_manageDebugMode.sqf
```diff
@@ -0,0 +1,23 @@
...
+
+if (_debugMode) then
+{
+    private _loopScriptHandle = [] spawn 
+        {
+            while { true } do
```
GermanHydrogen, line 7: Better use CBA_fnc_addPerFrameHandler, but I don't really understand the purpose of this loop

addons/main/XEH_preInit.sqf
```diff
@@ -1,2 +1,13 @@
...
\ No newline at end of file
+#include "XEH_PREP.hpp"
+
+[
+	"AE3_DebugMode", // Settings internal name
+	"CHECKBOX", // Settings type
+	["Debug Mode", "By enabling the AE3 Debug Mode you will get additional information about the internal structure of AE3."], // Settings Name + Tooltip
+	"AE3 main", // Settings category
+	false, // Default Value
+    nil, // "_isGlobal" flag. Set this to true to always have this setting synchronized between all clients in multiplayer
```
GermanHydrogen, line 10: Should always be forced by the server settings

GermanHydrogen (changes requested):

addons/armaos/functions/fnc_terminal_init.sqf
y0014984: Sorry, that's not the same.

`!isMultiplayer` means Singleplayer 
`isServer && hasInterface` means local hosted Multiplayer

In both cases not the CBA client settings but the CBA server settings are used

addons/main/functions/fnc_manageDebugMode.sqf
y0014984: With growing complexity it becomes more and more difficult to understand what's going on. Therefore I thought it would be a good idea to add a debug mode instead of using individual debug messages in the code. Then I thought we need triggers to execute these debug functions. A simple method is to check wether or not the debug mode is enabled directly in an individual debug function. Another useful way could be to trigger specific debug functions on interval. We will see in the future if this leads to something useful. 

One thing I already have in mind for a debug function is something I saw on lambs danger. The lambs danger debug mode adds an info box above every AI soldier. Something like that with basic information on power/network values could be useful for us.

addons/main/XEH_preInit.sqf
y0014984: Every CBA setting could be overwritten by mission or by server. There is no need to define that behavior at this point. If you think about using the `_isGlobal` flag, I think thats something different. This would sync the setting between all machines. We can't allow that because this would change the behavior for other players in a live mission as well.

GermanHydrogen (approved):

PR #203, opened 2022-11-11
Debug Overlay

y0014984:
This PR needs to wait until the PR #187 containing the Debug Mode CBA Setting is merged. Until then I will add additional information to the overlay. Probably we could also add lines to visualize established connections between devices.

- ACE cargo name
- device class
- power state
- connected devices count
- fuel level (3 values)
- power output
- battery level (3 values)

Also changed connectTo-interactions for power and network connections to show the ACE3 cargo name instead of device class. This is shorter and allows to use the ACE3 cargo rename feature so you are always sure to which device you establish the connection if there are multiple similar devices.

```sqf
// enable debug overlay
private _objects = missionNamespace getVariable "AE3_DebugOverlay";
[_objects] call AE3_main_fnc_initDebugOverlay;

//disable debug overlay
private _objects = missionNamespace getVariable "AE3_DebugOverlay";
[_objects] call AE3_main_fnc_killDebugOverlay;
```

Debug Overlay is now toggled via Debug Mode.


GermanHydrogen (approved):
Looks good. Additionally lines (drawLine3D) between connected devices would be nice.
Maybe with different colors for network/power. If you don't want to do this in this PR, I will submit this idea as a issue.

y0014984:
> Looks good. Additionally lines (drawLine3D) between connected devices would be nice.
> Maybe with different colors for network/power. If you don't want to do this in this PR, I will submit this idea as a issue.

I hab the same idea, too. But at the moment I would keep it in an issue.

PR #204, opened 2022-11-13
Standby fix

### LINKED ISSUE #186: Wakeup after Standby without authentication
If you use the `standby` command in the terminal and wake up the laptop again, you need to login again.

If you are logged in and close the laptop, the laptop goes to standby. If you open up the laptop again it boots. If you then use the terminal you are already logged in. This is a security issue and should be fixed.

GermanHydrogen:
Closes #186


y0014984 (approved):
Thank you. Looks good. Already tested in dedicated mp. Works as well. You can merhe this PR and release the final 0.5.0

PR #208, opened 2022-11-15
fixed a bug with 'exit' command and standby

y0014984:
@Wasserstoff removed this line of code in a previous fix. This breaks the `exit` command:

```sqf
[_computer] call AE3_armaos_fnc_terminal_addHeader;
```

Additionally there was a need to modify terminal init function for that case.


y0014984:
This should be a bugfix for the current release so perhaps we should merge into `master`?

GermanHydrogen (approved):

Issue #133, opened 2022-04-02
CTRL+C to cancel current Terminal input


y0014984:
CTRL+C to cancel current Terminal input. This isn't really needed for gameplay but I think this gives the real terminal feeling, if you could do that.

y0014984:
solved by @GermanHydrogen in v0.5.0

PR #217, opened 2022-11-18
Armaos terminal renderpipeline rework

GermanHydrogen:
Todo:

- [x] Input Buffer
- [x] Movable Coursor
- [x] Render Pipeline
- [x] Fix scrolling
- [x] Text Zoom

### DIFF
--- a/addons/armaos/XEH_PREP.hpp
+++ b/addons/armaos/XEH_PREP.hpp
@@ -53,6 +53,9 @@ PREP(terminal_removeCharFromInput);
 PREP(terminal_shiftInputBuffer);
 PREP(terminal_getInput);
 
+PREP(terminal_renderLine);
+PREP(terminal_reRenderBuffer);
+
 PREP(terminal_addChar);
 PREP(terminal_addLines);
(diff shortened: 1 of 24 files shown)


y0014984 (changes requested):
Had a quick code overview and tested the moveable cursor and the new zoom feature. Looks good to me. 

Please fix the login problem, I already mentioned. Additionally, could you please increase the used space at the bottom and the left side of the terminal? There is a lot of unused space in the different zoom levels. Also there seems to be additional blank lines at the login after login and password input. Could you remove them. Last but not least there are a few old code lines that are commented out. Are they still necessary?

y0014984:
Solving the ticket #133 would be a perfect addition to this pr, because now, if I edit very long lines and want to discard my changes, I need to delete the complete line or move down to the last history entry (which is empty). Using CTRL-C to break current input attempt would be nice in this situation. This is just a comment and not necessary to be included in this PR.

y0014984 (approved):
Looks good to me now. You can merge this PR.

PR #225, opened 2022-11-20
Localization (CHS) Chinese Simplified

PowerBOXx:


y0014984:
@PowerBOXx Thank you for chinese localization. There will be a lot of updates in the near future, so we need to update translations as well. I will keep you informed.

PR #228, opened 2022-11-21
Interaction fixes (Laptop, Desk, Single Lamp)

y0014984:


addons/interaction/functions/fnc_laptop_close.sqf
```diff
@@ -10,6 +10,8 @@
 
 params['_laptop'];
 
+_laptop enableSimulationGlobal false;
```
GermanHydrogen, line 13: I reviewed my recommondation and this does not work, because the simulation true state
is needed for the animation to work. So we can remove this.

addons/interaction/functions/fnc_laptop_open.sqf
```diff
@@ -10,6 +10,8 @@
 
 params['_laptop'];
 
+_laptop enableSimulationGlobal false;
```
GermanHydrogen, line 13: See note above.

addons/armaos/CfgVehicles.hpp
```diff
@@ -104,7 +104,7 @@ class CfgVehicles
 		{
 			displayName = "Laptop";
 
-			closeState = 1;
```
GermanHydrogen, line 107: The animation is still played, because the open fnc is played regardless of state.
There maybe should be a default closedState which is set according to the objects default animation state

GermanHydrogen (changes requested):

GermanHydrogen (approved):

PR #243, opened 2022-11-22
Interaction Restrictions

y0014984:
# Description
Merging this Pull Request will change the way ACE3 interactions like carry, drag and cargo are managed. All ACE3 interaction properties are still set in the config but now in the AE3 way. We use our on config that is parsed in the init phase like for all AE3 properties. A function will then enable/disable the ACE3 interactions dependig on the states of the device, see below.
All settings and states are held in a HashMap saved in the namespace of the device.

## States
- powerConnected = The device has an active power cable connection to one or more devices, no matter if it's providing or consuming power
- networkConnected = The device has an active network cable connection to one or more devices, no matter if it's a router or a computer
- unwrapped = Means that the device is in a state that does not allow transport, like the unfold desk. This is not strictly nesessary for the laptop.
- turnedOn = The device is powered on, which prevents for example transportation; This is not nessesarily the case for every device; I think you can carry a turned on laptop but it makes no sense for a turned on generator
- inUse = Means another player is currently using the device, like the laptop, so no other user should be able to carry the device around

## Benefits
- The presence of ACE3 interactions now depend on multiple states like power/network connections, device in use, device turned on etc. There can be added additional states easily
- By changing one of these states for a device, a function checks wether ACE3 interactions should be enabled or not; This depends on the result of all status types and is therefore very flexible
## Drawbacks
- The class/asset replace method for the generator had to be disabled because it leads to a script error in a ACE3 function dedicated to the ACE3 interaction framework (I don't know why); But the replace method also was not 100% accurate (deviced jumped approx. halb a meter to one side) so perhaps we should redesign it either way; Also the dummy classes behaved strangely when moved in eden editor (only yellow box moved, but modell stayed at position). **One solution could be to replace objects not on mission start but directly in eden editor, after placing them.**
- The generator is now unter blufor assets in the asset browser instead of props due to the prior problem

## How to use
If you set or change a state, you simply call this function with the corresponding state and set it to true or false. The function then enables or disables all ACE3 interactions depending on the status of all states combined. Here are some examples:
```sqf
[_computer, "inUse", true] remoteExecCall ["AE3_interaction_fnc_manageAce3Interactions", 2];
[_desk, "unwrapped", false] remoteExecCall ["AE3_interaction_fnc_manageAce3Interactions", 2];
[_lamp, "turnedOn", false] remoteExecCall ["AE3_interaction_fnc_manageAce3Interactions", 2];
[_device, "networkConnected", true] remoteExecCall ["AE3_interaction_fnc_manageAce3Interactions", 2];
[_target, "powerConnected", true] remoteExecCall ["AE3_interaction_fnc_manageAce3Interactions", 2];
```

## Config examples
```sqf
class AE3_Equipment
{
	class AE3_ace3Interactions
	{
		class AE3_aceDragging
		{
			// Dragging
			ae3_dragging_canDrag = 1;  // Can be dragged (0-no, 1-yes)
			ae3_dragging_dragPosition[] = {0, 1, 0};  // Offset of the model from the body while dragging (same as attachTo)
			ae3_dragging_dragDirection = 0;  // Model direction while dragging (same as setDir after attachTo)
		};
		class AE3_aceCarrying
		{
			// Carrying
			ae3_dragging_canCarry = 0;  // Can be dragged (0-no, 1-yes)
			ae3_dragging_carryPosition[] = {0, 1, 1};  // Offset of the model from the body while dragging (same as attachTo)
			ae3_dragging_carryDirection = 180;  // Model direction while dragging (same as setDir after attachTo)
		};
		class AE3_aceCargo
		{
			ae3_cargo_canLoad = 1;  // Enables the object to be loaded (1-yes, 0-no)
			ae3_cargo_size = 2;  // Cargo space the object takes
		};
		class AE3_interactionConditions
		{
			ae3_unwrapped = 1; // unwrapped = 1; wrapped = 0
		};
	};
};
```


y0014984:
Hi, I need your support on this PR. There a two isses (#258 and #259) I can't solve at the moment. Especially #258 is strange. I can't figure out why my PR created this bug, because I think I haven't changed anything related to this.

y0014984:
#259 is fixed now, #258 is still open.

y0014984:
#258 is now fixed too. @GermanHydrogen you can again review this PR.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+ * Example:
+ * [_computer, "powerConnected", true] call AE3_interaction_fnc_manageAce3Interactions
+ */
+
+params ["_target", "_condition", "_status"];
+
+// ace_dragging_fnc_setCarryable and ace_dragging_fnc_setDraggable need to be executed on all machines with JIP
+// ace_cargo_fnc_setSize can be executed locally because it handles JIP and remoteExecCall on it's own
+
+//this function only runs once per triggered interaction by player on players machine
```
GermanHydrogen, line 21: Why not run this server side only? All effects of this function are made globally anyway

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+        {
+            [_target, -1] call ace_cargo_fnc_setSize;
+        };
+    };
+
+    // ensures that cargo rename feature is enabled on all assets, no matter if canLoad is true
+    // some classes have a disabled renaming feature via ACE3 config, like the yellow lamps
+    _target setVariable ["ace_cargo_noRename", false, true];
+
+	_target setVariable ["AE3_SettingsACE3", _settingsAce3, true];
```
GermanHydrogen, line 95: Global variable sync should be avoided, espacially for hashmaps. Could be easility fixed, if this function is only run
server side. See above

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+params ["_target", "_condition", "_status"];
+
+// ace_dragging_fnc_setCarryable and ace_dragging_fnc_setDraggable need to be executed on all machines with JIP
+// ace_cargo_fnc_setSize can be executed locally because it handles JIP and remoteExecCall on it's own
+
+//this function only runs once per triggered interaction by player on players machine
+private _settingsAce3 = _target getVariable "AE3_SettingsACE3";
+if (!isNil "_settingsAce3") then
```
GermanHydrogen, line 23: Use of exitWith with negated condition would make the function cleaner.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+
+    private _powerConnected = _settingsAce3 getOrDefault ["powerConnected", false, true];
+    private _networkConnected = _settingsAce3 getOrDefault ["networkConnected", false, true];
+    private _unwrapped = _settingsAce3 getOrDefault ["unwrapped", false, true];
+    private _turnedOn = _settingsAce3 getOrDefault ["turnedOn", false, true];
+    private _inUse = _settingsAce3 getOrDefault ["inUse", false, true];
+
+    // ACE3 Carry, Drag and Load will only be available if the device is not connected via power/network, 
+    // not unfold/opened/expanded, not turned on
+    if (_powerConnected || _networkConnected || _unwrapped || _turnedOn || _inUse) then
```
GermanHydrogen, line 35: Wouldn't be it better, if this is only triggered, if the status is changed?
Concider the case, if a laptop is already connected to power, so everything is already disabled.
So this does not have to run, if it is then connected to a network.
This applies to the hole function.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+        if (_settingsAce3 get "ae3_cargo_canLoad") then
+        {
+            [_target, -1] call ace_cargo_fnc_setSize;
+        };
+    }
+    else
+    {
+        if (_settingsAce3 get "ae3_dragging_canDrag") then
+        {
+            private _canDrag = _settingsAce3 get "ae3_dragging_canDrag";
```
GermanHydrogen, line 54: This is always true

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+            private _dragDirection = _settingsAce3 get "ae3_dragging_dragDirection";
+            [_target, _canDrag, _dragPosition, _dragDirection] remoteExecCall ["ace_dragging_fnc_setDraggable", 0, true];
+        }
+        else
+        {
+            [_target, false] remoteExecCall ["ace_dragging_fnc_setDraggable", 0, true];
+        };
+        if (_settingsAce3 get "ae3_dragging_canCarry") then
+        {
+            private _canCarry = _settingsAce3 get "ae3_dragging_canCarry";
```
GermanHydrogen, line 65: This is always true

addons/interaction/functions/fnc_manageAce3Interactions.sqf
```diff
@@ -0,0 +1,96 @@
...
+            private _carryDirection = _settingsAce3 get "ae3_dragging_carryDirection";
+            [_target, _canCarry, _carryPosition, _carryDirection] remoteExecCall ["ace_dragging_fnc_setCarryable", 0, true];
+        }
+        else
+        {
+            [_target, false] remoteExecCall ["ace_dragging_fnc_setCarryable", 0, true];
+        };
+        if (_settingsAce3 get "ae3_cargo_canLoad") then
+        {
+            private _canLoad = _settingsAce3 get "ae3_cargo_canLoad";
```
GermanHydrogen, line 76: This is always true

addons/interaction/functions/fnc_initAce3Interactions.sqf
```diff
@@ -0,0 +1,71 @@
...
+ * 3: Cargo Settings <ARRAY>
+ *
+ * Returns:
+ * none
+ */
+
+params["_equipment", "_aceDragging", "_aceCarrying", "_aceCargo", "_interactionConditions"];
+
+// this function only runs once on mission start on server or on client if it's no multiplayer
+if(isServer || !isMultiplayer) then
```
GermanHydrogen, line 18: isServer is already true if called in singleplayer

addons/interaction/functions/fnc_initAce3Interactions.sqf
```diff
@@ -0,0 +1,71 @@
...
+	if (!(_interactionConditions isEqualTo [])) then
+	{
+		private _ae3_unwrapped = _interactionConditions select 0;
+
+		if (_ae3_unwrapped == 1) then { _ae3_unwrapped = true; } else { _ae3_unwrapped = false; };
+
+		_settingsAce3 set ["unwrapped", _ae3_unwrapped];
+	};
+
+	_equipment setVariable ["AE3_SettingsACE3", _settingsAce3, true];
```
GermanHydrogen, line 68: If managedAce3Interactions is calles server side, the global flag is unnecessary

GermanHydrogen (changes requested):
The system in general looks solid. Only the function manageAce3Interactions needs major optimization.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: Ok, I will try to change that. But I think the only benefit is that this chunk of code (that already only runs once) is then executed on the server instead of the client. I think I will simply change that by remote executing this function on the server.

```sqf
// doing this
remoteExecCall 
[_computer, "inUse", false] remoteExecCall ["AE3_interaction_fnc_manageAce3Interactions", 2];

// instead of this
[_computer, "inUse", false] call AE3_interaction_fnc_manageAce3Interactions;
```

addons/interaction/functions/fnc_initAce3Interactions.sqf
y0014984: You are right. I will remove `|| !isMultiplayer`

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: Will change that.

addons/interaction/functions/fnc_initAce3Interactions.sqf
y0014984: Will change that.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: Will do that.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: I will introduce another veriable holding the current `_isBlocked` status. So blocking and unblocking is only done if it's a change of state.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: Now I know what you mean. If I'm already checking the var in the if-statement, I do not need to read and save that var again, as I already know the content of the var. I will fix that.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: Now I know what you mean. If I'm already checking the var in the if-statement, I do not need to read and save that var again, as I already know the content of the var. I will fix that.

addons/interaction/functions/fnc_manageAce3Interactions.sqf
y0014984: Now I know what you mean. If I'm already checking the var in the if-statement, I do not need to read and save that var again, as I already know the content of the var. I will fix that.

y0014984:
Hi @GermanHydrogen , I think I have managed to address all your comments. Feel free to review my changes.

GermanHydrogen (approved):
Looks good

PR #256, opened 2022-11-27
Fixed exception handling for createFile and createDir function calls

y0014984:
By merging this Pull Request every call of createDir or createFile function is embedded in a try/catch-block. Both functions thow errors if the filesystem object already exists. This happens for example if you snyc identical modules multiple times to the same computer. The exceptions are now logged in the *.rpt file.

### DIFF
--- a/addons/filesystem/functions/fnc_initFilesystem.sqf
+++ b/addons/filesystem/functions/fnc_initFilesystem.sqf
@@ -35,26 +35,34 @@ private _filesystemObjects = ("inheritsFrom _x == (configFile >> 'AE3_Filesystem
 
 	if (_type isEqualTo "File") then
 	{
-		[
-			_ptr, 
-			_filesystem, 
-			_path, 
-			(getText (_x >> "content")), 
-			"root", 
-			_owner,
-			_permissions 
-		] call AE3_filesystem_fnc_createFile;
+		// throws exception if file already exists
+		try 
+		{
+			[
+				_ptr, 
+				_filesystem, 
+				_path, 
+				(getText (_x >> "content")), 
+				"root", 
+				_owner,
+				_permissions 
+			] call AE3_filesystem_fnc_createFile;
+		} catch { diag_log format ["AE3 exception: %1", _exception]; };
 	}
 	else
 	{
-		[
-			_ptr,
-			_filesystem,
-			_path,
-			"root",
-			_owner, 
-			_permissions
-		] call AE3_filesystem_fnc_createDir;
+		// throws exception if directory already exists
+		try 
+		{
+			[
+				_ptr,
+				_filesystem,
+				_path,
+				"root",
+				_owner, 
+				_permissions
+			] call AE3_filesystem_fnc_createDir;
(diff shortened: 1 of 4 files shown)


addons/filesystem/functions/fnc_initFilesystem.sqf
```diff
@@ -35,26 +35,34 @@ private _filesystemObjects = ("inheritsFrom _x == (configFile >> 'AE3_Filesystem
...
+			[
+				_ptr, 
+				_filesystem, 
+				_path, 
+				(getText (_x >> "content")), 
+				"root", 
+				_owner,
+				_permissions 
+			] call AE3_filesystem_fnc_createFile;
+		} catch { diag_log format ["AE3 exception: %1", _exception]; };
```
GermanHydrogen, line 50: To broad catch. Would be better, if other exceptions are still raised

addons/filesystem/functions/fnc_initFilesystem.sqf
```diff
@@ -35,26 +35,34 @@ private _filesystemObjects = ("inheritsFrom _x == (configFile >> 'AE3_Filesystem
...
+		{
+			[
+				_ptr,
+				_filesystem,
+				_path,
+				"root",
+				_owner, 
+				_permissions
+			] call AE3_filesystem_fnc_createDir;
+		} catch { diag_log format ["AE3 exception: %1", _exception]; };
```
GermanHydrogen, line 65: See above

addons/filesystem/functions/fnc_moduleAddDir.sqf
```diff
@@ -40,14 +40,19 @@ if(_path isEqualTo "") exitWith {};
...
+		{
+			[
+				[],
+				_filesystem,
+				_path,
+				"root",
+				_owner,
+				_permissions
+			] call AE3_filesystem_fnc_createDir;
+		} catch { diag_log format ["AE3 exception: %1", _exception]; };
```
GermanHydrogen, line 55: See above

addons/filesystem/functions/fnc_moduleAddFile.sqf
```diff
@@ -47,15 +47,21 @@ if(_isFunction) then
...
+			[
+				[],
+				_filesystem,
+				_path,
+				_content,
+				"root",
+				_owner,
+				_permissions
+			] call AE3_filesystem_fnc_createFile;
+		} catch { diag_log format ["AE3 exception: %1", _exception]; };
```
GermanHydrogen, line 63: see above

GermanHydrogen (changes requested):

y0014984:
@GermanHydrogen ,I modified the catch clauses to only catch 'already exists' exceptions, so other exceptions are still thrown. It was a bit complicated because the exception is not a fixed string but contains the name of the filesystem object that already exists. So I needed to regExReplace the dynamic content for example `'home' already exists!` to look like this `'%1' already exists!`. This this the `format` string, that is used to build the dynamic exception. So, I can use this to compare the strings in the `if` clause.

GermanHydrogen (approved):

PR #260, opened 2022-12-05
split texture variants into separate classes

y0014984:
Merging this pull request will revert AE3 class creation to directly inherit from origin classes without overwriting class properties. We've done that in the past to allow changing the texture of placed asset in eden editor, which can be done via the change vehicle texture function. But this comes with some drawbacks, foremost some assets behaved strangly. For example the single rugged lamp stayed in mid-air after lamp was carried instead of falling back to ground (default behavior). Merging this pull request will fix that issue.


GermanHydrogen (approved):
Looks good

PR #261, opened 2022-12-05
Init-fuel-battery-levels

y0014984:
At the moment we have input fields on all editor assets to adjust the inital power and fuel levels. But these settings are ignored at the moment. Also the class setting for fuel level is ignored. 

Merging this Pull Request, both eden attributes can be used to define initial power and fuel levels. Unfortunately this does not work as excepted. We can't set fuel levels in in the init phase. Doesn't seem to work. So I added a postInit function to process that. But that doesn't work either. The only thing that works is setting battery level for laptop, but only if there is a delay of approx. 5 sec. on misson startup. All other assets (batteries, solar panels, generators) only work when I execute the same function in-game via the debug console. This is very strange.


y0014984:
@GermanHydrogen I also need your help on this on.

GermanHydrogen (commented):
Looks good, but it would be nice, if the attributes would only appear for the relevant objects

y0014984:
@GermanHydrogen You can review this PR again. Now the attributes only show up on the assets that need them. They now appear as a custom attribute. We can't define a category name for them. Also they replace any prior custom attributes but I think that's not a problem but an advantage, because setting the texture of the laptop in the editor or the angles of the solar panel in the editor is now not possible animore.

Also I changed the attributes from text input fields to sliders. I think this is more convenient.

GermanHydrogen (approved):
Looks good

PR #262, opened 2022-12-05
changed history to user dependent history

y0014984:
At the moment our history is not user dependent. All users share one history. That's a security issue. By merging this pull request a separate history is saved for every user. It's no more an Array but a HashMap of Arrays with the username as the key. I have successfully tested this in dedicated multiplayer.


GermanHydrogen (approved):
Looks good

PR #264, opened 2022-12-05
Critical bug fix for config parsing

GermanHydrogen:
Fixed config existence check by replacing isNull through !isClass
Fixed config buffer key by adding '_power' to key. (Avading possible collision)


y0014984:
@GermanHydrogen : I understand the problem now. We set a missionNamespace variable with the class name as var name for every class present in a mission. If it's a AE3 class the content of the var is our class config. If it's another class the content of the var is "". Unfortunately other mods do something like that as well. So, if the execution was in a unfortunate order, we have overwritten these variables set by other mods with our "" content. This breaks potentially other mods, as in our case the ACE tactical ladder. You can merge this pull request.

PR #290, opened 2022-12-13
Turn on/off progress bar cancellation for generator and solar panel

y0014984:
At the moment if the ACE3 progress bar for turning on/off generator or solar panel is cancelled, the setup is in an undefined state because some code like setting the on/off state is set although the action was cancelled.

By merging this pull request this issue is fixed. Now It's possible to cancel the progress bars. It's not a very elegant solution because we are switching from unscheduled to scheduled execution and therefore the handling of the return value is a problem. But I think the workaround is ok.

Another solution would be getting rid the progress bar at all. I don't think that the solar panel needs a progress bar. It could be turned on/off immediately. Also the generator doesn't need a interruptable stop/stop procedure. I think it's more like pressing a button. The start and stop sounds are only for immersion.

### DIFF
--- a/addons/power/functions/fnc_turnOffGeneratorAction.sqf
+++ b/addons/power/functions/fnc_turnOffGeneratorAction.sqf
@@ -67,7 +67,7 @@ else
 {
 	[_entity] call _turnOffGenFunc;
 
-	_return = true;
+	_result = true;
 };
 
 // function immediately returns false, because progress bar runs unscheduled
--- a/addons/power/functions/fnc_turnOffGeneratorAction.sqf
+++ b/addons/power/functions/fnc_turnOffGeneratorAction.sqf
@@ -49,7 +49,7 @@ if (!_silent) then
 			_entity setVariable ["AE3_power_powerState", 0, true];
 		},
 		{
-			// following code only runs on progress bar success
+			// following code only runs on progress bar fail
 			params ["_args", "_elapsedTime", "_totalTime", "_errorCode"];
 			
 			_args params ["_entity", "_stopSoundHandle", "_turnOffGenFunc"];
--- a/addons/power/functions/fnc_turnOffGeneratorAction.sqf
+++ b/addons/power/functions/fnc_turnOffGeneratorAction.sqf
@@ -8,35 +8,67 @@
  * Return:
  * None
 */
+
 params ["_entity", ["_silent", false]];
 
-_powerState = _entity getVariable "AE3_power_powerState";
+private _result = false;
 
-_turnOffTime = 3;
+private _turnOffTime = 3;
 
-_handle = [_entity] spawn AE3_power_fnc_playGeneratorStopSound;
+private _stopSoundHandle = [_entity] spawn AE3_power_fnc_playGeneratorStopSound;
 
+private _turnOffGenFunc =
+{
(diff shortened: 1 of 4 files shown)


addons/power/functions/fnc_turnOffGeneratorAction.sqf
```diff
@@ -8,35 +8,67 @@
...
-};
+			[_entity] call _turnOffGenFunc;
 
-[_entity, true, [0, 1, 0], 0] call ace_dragging_fnc_setDraggable;
+			// we need to set power state here because function already returned false
+			// and therefore the turn on wrapper doesn't set the state to turned on
+			_entity setVariable ["AE3_power_powerState", 0, true];
+		},
+		{
+			// following code only runs on progress bar success
```
GermanHydrogen, line 52: do you mean "on progress bar failure"?

addons/power/functions/fnc_turnOffSolarAction.sqf
```diff
@@ -8,30 +8,61 @@
...
+			[_entity] call _turnOffSolPanFunc;
 
+			// we need to set power state here because function already returned false
+			// and therefore the turn on wrapper doesn't set the state to turned on
+			_entity setVariable ["AE3_power_powerState", 0, true];
 		},
-		{},
-		(localize "STR_AE3_Power_Interaction_TurnOff")
+		{
+			// following code only runs on progress bar success
```
GermanHydrogen, line 52: same as above.

addons/power/functions/fnc_turnOnSolarAction.sqf
```diff
@@ -2,28 +2,43 @@
...
+		[_entity, AE3_power_fnc_solarCalculation] remoteExecCall ["AE3_power_fnc_addProviderHandler", 2];
+
+		[_entity, false, [0, 1, 0], 0] call ace_dragging_fnc_setDraggable;
+
+		// we need to set power state here because function already returned false
+		// and therefore the turn on wrapper doesn't set the state to turned on
+		_entity setVariable ["AE3_power_powerState", 1, true];
+	},
+	{
+		// following code only runs on progress bar fail
```
GermanHydrogen, line 35: could be empty

addons/power/functions/fnc_turnOffGeneratorAction.sqf
GermanHydrogen: could be empty

GermanHydrogen (changes requested):

addons/power/functions/fnc_turnOffGeneratorAction.sqf
y0014984: Yes, you are right. I will fix that.

addons/power/functions/fnc_turnOffSolarAction.sqf
y0014984: See above.

addons/power/functions/fnc_turnOnSolarAction.sqf
y0014984: Yes, I think I will remove the unnecessary code lines.

GermanHydrogen (approved):

PR #293, opened 2022-12-14
Disable empty login string in debug mode and apply filter

y0014984:
Prior to this pull request logins in debug mode with empty strings or strings containing blanks and special characters are allowed.

Merging this Pull Request will add a check for empty strings and applies a filter (a..z, A..Z, 0..9, "_") to debug mode usernames, but only if the entered username is not a legal username in this particular armaOS instance.


GermanHydrogen (approved):
Looks good

PR #294, opened 2022-12-14
Fixed armaOS use delay

y0014984:
If you bootup a laptop and the startup sound appears there is still a 5 sec. delay, before the ACE3 Interaction to use armaOS appears. This is because the turnOn function runs in scheduled environment. So every called function also runs in scheduled environment. Therefore the `sleep 5;` in the playSoundStart function has an effect and pauses the execution. By merging this Pull Request we get rid of that command, because I see no need for that.


GermanHydrogen (approved):
Looks good

PR #295, opened 2022-12-16
3DEN Editor custom power and network connections

y0014984:
Merging this Pull Request will add custom Eden Editor connections for power and network, see screenshots. This allows the mission creator to connect devices bevore mission start.

It's very important, that these connections are established after all devices are fully initialized. Therefore I needed to introduce a new variable.

### Features
- Power and Network connections are only allowed between objects (not triggers, markers etc.) otherwise removed
- Power and Network connections use different colors to be distinguishable (colors are only shown for active device)
- After an established connection there are checks if the connection is allowed (class and direction based) otherwise removed mit an eden editor hint

**Keep in Mind:** Router to router connections behave weirdly at the moment as long as there is no NAT. So the ip network is shared acroos all routers and connected devices. But this only works if the master router is turned on first. We should optimize that.


addons/main/functions/fnc_3denEventHandlers_onConnectionEnd.sqf
```diff
@@ -0,0 +1,112 @@
...
+    {
+        ["AE3_NetworkConnection", _x, _to, _allowedNetworkFromClasses, _allowedNetworkToClasses] call AE3_main_fnc_3den_checkConnection;
+    } forEach _fromObjects;
+};
+
+/* ================================================================================ */
+
+if (_type isEqualTo "AE3_PowerConnection") then
+{
+    private _allowedPowerFromClasses =
```
GermanHydrogen, line 61: It would be better, if we check the config for `AE3_Generator`, `AE3_Battery` and `AE3_Consumer`, instead
of hardcoding the types.

addons/main/functions/fnc_3denEventHandlers_onConnectionEnd.sqf
```diff
@@ -0,0 +1,112 @@
...
+
+        //(format ['_object: %1', _object]) call BIS_fnc_3DENNotification;
+    }
+]; */
+
+/* ================================================================================ */
+
+if (_type isEqualTo "AE3_NetworkConnection") then
+{
+    private _allowedNetworkFromClasses =
```
GermanHydrogen, line 34: In the future we should introduce a network config in the CfgVehicles of network devices, so we 
dont have to hard code the types here.

addons/main/functions/fnc_3denEventHandlers_onConnectionStart.sqf
```diff
@@ -0,0 +1,9 @@
+params ["_type", "_from"];
+
```
GermanHydrogen, line 2: Empty file?

addons/main/functions/fnc_3den_checkConnection.sqf
```diff
@@ -0,0 +1,9 @@
+params ["_type", "_from", "_to", "_allowedFromClasses", "_allowedToClasses"];
+
+if (!((typeOf _from) in _allowedFromClasses) || !((typeOf _to) in _allowedToClasses) || (_from isEqualTo _to)) then
+{
+    remove3DENConnection [_type, [_from], _to];
+    "forbidden connection removed" call BIS_fnc_3DENNotification;
```
GermanHydrogen, line 6: Is there a way to explain the issue better?
Because the vanilla connections are symetrical, our approche could be unintuitive

addons/main/functions/fnc_3den_doNetworkConnection.sqf
```diff
@@ -0,0 +1,25 @@
...
+
+// _entity0 = from object
+// _entity1 = to object
+
+//diag_log [_type, _entity0, _entity1];
+
+[_entity0, _entity1] spawn 
+{
+    params ["_entity0", "_entity1"];
+    waitUntil { !isNil "BIS_fnc_init" };
```
GermanHydrogen, line 11: Probably unnecessary if we wait for the variables to be set?

addons/main/functions/fnc_3den_doPowerConnection.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+// _entity0 = from object
+// _entity1 = to object
+
+//diag_log [_type, _entity0, _entity1];
+
+[_entity0, _entity1] spawn
+{
+    params ["_entity0", "_entity1"];
+    waitUntil { !isNil "BIS_fnc_init" };
```
GermanHydrogen, line 11: See above

GermanHydrogen (approved):

addons/main/functions/fnc_3denEventHandlers_onConnectionStart.sqf
y0014984: This file is only there to tell you, that it's not necessary. But I think we can remove it. I will do that.

addons/main/functions/fnc_3den_checkConnection.sqf
y0014984: In vanilla all connections are allowed. You can sync averything to everything. I never liked that because the user thinks: "If I can do that, then it's valid and will work." But that's not always the case. With these distinct connection types for network and power, It should only be possible to connect things together that are meant to be connected. We could explain more in detail, if I split into separate error messages. But we should still remove connections that are wrong, for example a network connection between desk and chair. I will spit the error messages to be more precise and informative.

addons/main/functions/fnc_3den_doNetworkConnection.sqf
y0014984: Yes, you are right. I will remove that.

addons/main/functions/fnc_3den_doPowerConnection.sqf
y0014984: Yes, will be removed.

addons/main/functions/fnc_3denEventHandlers_onConnectionEnd.sqf
y0014984: Yes you are right. For the moment I will leave this as it is and create an issue for that.

addons/main/functions/fnc_3denEventHandlers_onConnectionEnd.sqf
y0014984: Yes, you are right. I found a solution for that.

GermanHydrogen (approved):
Looks good

PR #296, opened 2022-12-25
Retro Games using pixel based canvas

y0014984:
## Misc
- [x] Sync module to computer to allow games

## Functions
- [x] createCanvas
- [x] setPixelColor
- [x] getPixelColor
- [x] getCanvasWidth
- [x] getCanvasHeight
- [x] showSplashScreen

## Games
- [x] snake


y0014984:
@GermanHydrogen You already can have a look at this pull request, if you are interested. The core seems to be useable and `snake` is fun. From now on I will only add useful 'game engine' commands and additional games.

GermanHydrogen:
> @GermanHydrogen You already can have a look at this pull request, if you are interested. The core seems to be useable and `snake` is fun. From now on I will only add useful 'game engine' commands and additional games.

I think the games should not be on the OS per default. We could use a module to let the mission builder install them.

y0014984:
> I think the games should not be on the OS per default. We could use a module to let the mission builder install them.

I knew that you would say that. Therefaore I already added a Module. Only if the Module is synced to the computer, the games will show up.

y0014984:
It's to much to create all games now. So please review as it is. Additional games and function will get into that later.

y0014984:
Last things to do:
[ ] inherit commands from osCommand instead of introducing new classes

y0014984:
This Pull Request will be discontinued and the branch will be deleted due to mistakes with branch rebase. The new branch and Pull Request can be found here #317

PR #297, opened 2022-12-27
Terminal Redesign

y0014984:
Merging this pull request will update the UI design, so it reflects the static textures on the laptop. Also, all buttons are now on the upper right corner and can be used to change the design, change keyboard layout, check battery status and leave armaOS. The user can also change the design within the armaOS CBA settings. Actually, there are 4 different designs to choose from (default, C64, Apple II, Amber, see screenshots). Also the terminal window is now moveable. This will come in handy if we will allow additonal windows to show for example media or a retro gaming canvas.

- [x] new terminal design like on static textures
- [x] switching between multiple terminal designs via button
- [x] move all buttons to upper right corner
- [x] make the window movable by dragging title bar
- [x] added CBA setting for terminal designs
- [x] pre-created localisation of CBA setting
- [x] get battery status hint by clicking on battery symbol
- [x] new icons for closing and change design
- [x] updated default textures 
- [x] fixed shutdown texture order (reverse progressbar is correct)

AE3_Terminal_Designs


GermanHydrogen (commented):
Would probably be easier to implement the terminal for PC's if the battery symbol would be left
of the language selector.

y0014984:
> Would probably be easier to implement the terminal for PC's if the battery symbol would be left
> of the language selector.

Yes. You are right. I will change that.

GermanHydrogen (approved):
Looks good

PR #298, opened 2022-12-28
armaOS accessable while sitting on chair

y0014984:
Merging this PR will change the ACE Interaction for the laptop. It's now defined in init script and not in config. This allows the player to access armaOS while sitting on a chair with ACE3 sitting. The creation of the action via config seems to prevent that. I don't know why but this fixes the issue.

Now the ACE3 interactions in all places are all defined via script and not via config.


GermanHydrogen (approved):
Looks good

PR #300, opened 2022-12-30
added modules 'addSecurityCommands' and 'addHackingCommands'

y0014984:


addons/armaos/functions/fnc_hackingCommands_init.sqf
```diff
@@ -0,0 +1,32 @@
+/**
+ * Initilizes all hacking commands defined in CfgHackingCommands.
```
GermanHydrogen, line 2: Probably can be abstracted, to work for all command related modules

addons/armaos/functions/fnc_module_addHackingCommands.sqf
```diff
@@ -0,0 +1,17 @@
+params["_logic", "_units", "_activated"];
```
GermanHydrogen, line 1: Probably can be abstracted, to work for all command related modules

addons/armaos/functions/fnc_module_addSecurityCommands.sqf
```diff
@@ -0,0 +1,17 @@
+params["_logic", "_units", "_activated"];
```
GermanHydrogen, line 1: Probably can be abstracted, to work for all command related modules

addons/armaos/functions/fnc_securityCommands_init.sqf
```diff
@@ -0,0 +1,32 @@
+/**
```
GermanHydrogen, line 1: Probably can be abstracted, to work for all command related modules

addons/armaos/CfgHackingCommands.hpp
```diff
@@ -0,0 +1,20 @@
+class HackingCommand
```
GermanHydrogen, line 1: Is it possible, that all commands are inherited by the same parent (OsFunction)?

GermanHydrogen (changes requested):
Would be nice, if one could choose the commands added by a module with a checkboxes.
If this is out of scope, please add as issue

addons/armaos/CfgHackingCommands.hpp
y0014984: I asked myself the same question. Because I need to filter for specific classes, it was easier to introduce completely new classes. But I could do something like inheriting from `OsFunction` or I could add an additional value/parameter for filter purposes only. I need to think about that.

y0014984:
> Would be nice, if one could choose the commands added by a module with a checkboxes.

Interesting Idea. I will try to implement that.

addons/armaos/functions/fnc_hackingCommands_init.sqf
y0014984: Yes, at the moment it produces a lot of overhead. I need a distinguishable feature to reduce the amount of code. I will think about that.

y0014984:
Last things to to:

- [ ] change classes for commands back to OsCommand

addons/armaos/CfgHackingCommands.hpp
y0014984: Will do that.

y0014984:
@GermanHydrogen , you can review my latestest changes now. I consolidated the two modules into one. Inside the module you can choose the commands to sync by setting/removing the checkboxes. Also I consolidated the functions to add the commands to armaOS.

GermanHydrogen (approved):
Looks good

PR #303, opened 2023-01-03
Damage for props (WIP)

### LINKED ISSUE #25: destroyable equipment, so it becomes useless
For now I do not know what happens if we destroy some of the assets.

- Do every asset has a damage asset?
- Is damage tracked for every asset?
- what should happen if the asset is destroyed?

We should define and test that. Every item should be destroyable. If there is no damage model we should despawn the asset, like it's blown to dust. Every functionality should stop at this point.

y0014984:
# TODO

- [ ] Bug: destroyedWrapper has wront content
- [ ] change laptop texture while damaged (see terminal overlay)
- [ ] add damage config to all assets
- [ ] add damaged/destroyed behavior to all assets


A3-Root:
This has been abandoned.

PR #304, opened 2023-01-07
Armaos flash drive

GermanHydrogen:
Flash Drives
- [x] Flash Drive Item Icon
- [x] Mount & Unmount commands
- [x] Interface framework
- [x] Unmount on take
- [x] Localization
- [x] Variable synchronization


y0014984:
Next I will take a look at your code and comment it like usual.

addons/armaos/CfgOsFunctions.hpp
```diff
@@ -175,4 +175,18 @@ class CfgOsFunctions
 		man = "$STR_AE3_ArmaOS_Config_CommandFindMan";
 		code = "_this call AE3_armaos_fnc_os_find";
 	};
+	class mount : OsFunction
+	{
+		path = "/bin/mount";
+		description = "$STR_AE3_ArmaOS_Config_CommandFindDescr";
```
y0014984, line 181: "description" and "man" are both taken from the `find` command. It's the same for `mount` and `unmount`.

addons/armaos/functions/fnc_os_mount.sqf
```diff
@@ -0,0 +1,26 @@
...
+ * 1: Computer <OBJECT>
+ * 2: File <[STRING]>
+ *
+ * Results:
+ * None
+ */
+
+params ["_computer", "_options"];
+
+if (count _options > 1) exitWith { [ _computer, format [localize "STR_AE3_ArmaOS_Exception_CommandHasTooManyOptions", "mount"] ] call AE3_armaos_fnc_shell_stdout; };
```
y0014984, line 14: We should keep in mind to later convert this part to the new getOpts workflow.

addons/filesystem/functions/fnc_chown.sqf
```diff
@@ -0,0 +1,33 @@
+params['_pntr', '_filesystem', '_target', '_user', '_owner', ['_recursive', false]];
```
y0014984, line 1: If there is a `chown` filesystem command, we should allow this command as an os command, too.

addons/flashdrive/CfgVehicles.hpp
```diff
@@ -0,0 +1,29 @@
+class CfgVehicles 
+{
+	/* ================================================================================ */
+	// Flash Drive
+	class Land_USB_Dongle_01_F;
+	class FlashDrive_AE3: Land_USB_Dongle_01_F
```
y0014984, line 6: We should keep the original class name and the add "_AE3", like in all other cases.

addons/flashdrive/functions/fnc_compileDevice.sqf
```diff
@@ -0,0 +1,36 @@
+/**
+ * Compiles the interface from config for an object's class.
```
y0014984, line 2: interface?

addons/flashdrive/stringtable.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Project name="AE3">
+    <Package name="FlashDrive">
+        <Container name="Interaction">
+            <Key ID="STR_AE3_Flashdrive_Interaction_Connect">
+                <Original>Connect Flash Drive</Original>
```
y0014984, line 6: Genrally speaking, please add all currently supported languages to the stringtable and fill in the english version. This is easier for translators.

GermanHydrogen: English is filled in?
How am I supposed to translate into languages, I don't know?

y0014984 (commented):
I do not fully understand everything you've done. Therefore, I would need to dig very deep into the code. That's not necessary at the moment, I would say. I assume the code works. That's what the tests have shown.

addons/flashdrive/stringtable.xml
y0014984: What I am trying to say is: Add the tags for other languages likes russian, but fill in the english texts into the russian tags, for example. So they are prefilled and can easily be translated.

y0014984: Translators often make errors fill in the tags themselves and create errors when they do so. Finding the errors is very time consuming. I already had that case.

addons/flashdrive/functions/fnc_compileDevice.sqf
y0014984: Forget this point. With the word "interface" I thought of an user interface but you meant USB interface.

addons/filesystem/functions/fnc_chown.sqf
GermanHydrogen: I will do this, when getOpts is merged

addons/armaos/functions/fnc_os_mount.sqf
GermanHydrogen: I will do this, when getOpts is merged

GermanHydrogen:
> Script Error moving a file from laptop to stick, see screenshot
> Script Error moving a file from stick to laptop, see screenshot

I'm pretty shure, these are bugs in mvObj code

> We can't add to or create files on the sticks. That limits a lot the possibilities. We really should add the cp command to allow copying files onto and from the usb stick. Perhaps we can add the output redirection via > and >> soon. This would be another solution.

In theory this is already implemented in mvObj. There is only an os command missing.

> Script Error accessing ACE3 Interaction menu of laptop, see screenshot (only in dedicated mp)

I could only reproduce this in editor mp.

> I would change USB0, USB1 etc. to lowercase. Everything else is lowercase, too. This was my first mistake trying to mount with lowercase. The missing man page lead to gueswork.

I would like to introduce a command, to list the available interfaces.

y0014984:
> I would like to introduce a command, to list the available interfaces.

Sounds like a good idea.

addons/armaos/CfgOsFunctions.hpp
```diff
@@ -175,4 +175,32 @@ class CfgOsFunctions
...
+	{
+		path = "/bin/lsusb";
+		description = "$STR_AE3_ArmaOS_Config_CommandLsusbDescr";
+		man = "$STR_AE3_ArmaOS_Config_CommandLsusbMan";
+		code = "_this call AE3_armaos_fnc_os_lsusb";
+	};
+	class chown : OsFunction
+	{
+		path = "/bin/chown";
+		description = "$STR_AE3_ArmaOS_Config_CommandUmountDescr";
```
y0014984, line 202: 'description' and 'man' are taken from `umount` instead of `chown`

y0014984 (changes requested):
I've tested your PR in dedicated mp. 

What looks good:
- 'chown' command (except the man and description entries)
- 'lsusb' command
- getOpts for 'mount and 'umount'
- usb stick numbers (except the bug, see below)
- 'mv' command for moving files from/onto usb stick

What needs adjustments/fixes:
- There is still an error showing up, see screenshot, if I access the ACE3 Interaction menu. It seems to be related to the suspending in the getRemoteVar, probably used by a condition of the menu entries.
- Man and Description entries for 'chown'
- If I use multiple usb sticks, they seem to be rearranged while attached to the latop. Steps to reproduce:
   - attach stick 1 to usb0
   - attach stick 2 to usb1
   - remove (visually chosen) stick 2
   - results in stick 1 in players inventory
- Also there is a renumbering bug. I used 4 usb sticks. The sticks mentioned earlier are stick 1 and 2. Then I joined with another player and took the 2 additional sticks. They identified themselfes also as stick 1 and two. As I switched back to the first player, the 2 sticks still connected to the laptop where no more numbered 1 and 2, but 3 and 4. This renumbering is irritating.

I will now work on the preview image of the usb stick. I will send you results as direct message.

y0014984 (approved):
Latest changes look good to me.

Issue #278, opened 2022-12-10
Set ACE3 Cargo name while in Eden Editor


y0014984:
Withe the ACE3 Cargo names it's easy to distinguish between multiple devices of the same class/type, but you need to rename the devices on mission start. Perhaps we can allow to set the ACE3 cargo names vie Eden Editor Attributes so the mission creator can already give names to the devices.

PR #305, opened 2023-01-15
Add 3DEN Editor attribute for ACE Cargo Name

indig0fox:
Entered name will appear in ACE Cargo List appended to display name of object.
"Laptop (Olive) [Laptop 1]"
See https://github.com/acemod/ACE3/blob/master/addons/cargo/functions/fnc_getNameItem.sqf

Please test/adjust :)


indig0fox:
Related to #278

y0014984:
Thank you, @indig0fox 

I think I will have a look into it on tuesday.

y0014984 (changes requested):
Hi @indig0fox , are you sure that this is working? I tested your PR, but the custom text field does not show up. Perhaps you missed the `displayName` attribute for the `AE_Properties` class, see link below. Or it's because you can't add properties to specific classes this way but only to all classes, the also the example. You nee to replace the Classname with "Object". Or perhaps I have done something wrong while testing.

Additionally, I should mention a few things.

- If we use an abbreviation, then it's AE3, not AE So it should be AE3_Properties, not AE_Properties. 
- We changed our custom eden attributes to "Entity Specific" attributes, see here: https://community.bistudio.com/wiki/Eden_Editor:_Configuring_Attributes You could change this attribute to to fit in that pattern
- We need this attribute for all AE3 classes, not only for the Laptop; You find them in multiple addons
- Coincidentally I saw that you can change the cargo name within the ACE3 specific Eden Attributes, but only for vehicles. So this only affects our generator at the moment, but not the other assets; Perhaps we could inject this field into the ACE3 properties, or make a PR at ACE3 to implement this field into all assets, not only vehicles.
- Also we should add localisazion strings
- The default value is perhaps no good choice. If you place 3 laptops in the editor and don't change the names, you will end up with 3 times "Laptop 1". We should leave this property at ""

Will you change this PR? Do you have any additional questions?

I am happy that there are additional contributors. Sorry, that I can't merge the PR in this state.

indig0fox:
All good! First time contributing to the project and wanted to lay the framework for how this could be accomplished as I've done similar before -- with your constructive feedback I can proceed further and align it to your vision. I'll convert this to a draft until it's ready.

Issue #116, opened 2022-04-02
ACE3 Progress Bar for unfolding/folding rugged desk


y0014984:
Unfolding/folding the desk should be a time consuming task. Therefore the animation should take some time. We should add an ACE3 rpogress bar for this animation like we do for other assets/animations. 10 to 15 seconds should be enough.

PR #306, opened 2023-01-15
ACE3 Progress Bar for opening/closing rugged desks

indig0fox:
#116
https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_progressBar.sqf

Changes open/close action of AE desks to enter start-open/start-close wrapper that provides progressBar. Customize conditions as desired

### DIFF
--- a/addons/interaction/functions/fnc_start_desk_close.sqf
+++ b/addons/interaction/functions/fnc_start_desk_close.sqf
@@ -1,0 +1,23 @@
+/**
+ * Starts process of closing a desk
+ *
+ * Arguments:
+ * 0: Desk <OBJECT>
+ *
+ * Return Value:
+ * None
+*/
+
+params ['_desk'];
+
+[
+	15, // time in seconds
+	_desk, // arguments for Finish + fail
+	{ // on finish
+		[_this] call AE3_interaction_fnc_desk_close;
+	},
+	{}, // on fail
+	"Closing desk" // progress bar text, can be localized
+	{true}, // condition code checked each frame to continue
+	["isNotSwimming"] // conditions for checking interaction
+] call ace_common_fnc_progressBar;
--- a/addons/interaction/functions/fnc_start_desk_open.sqf
+++ b/addons/interaction/functions/fnc_start_desk_open.sqf
@@ -1,0 +1,23 @@
+/**
+ * Starts process of opening a desk
+ *
+ * Arguments:
+ * 0: Desk <OBJECT>
+ *
+ * Return Value:
+ * None
+*/
+
+params ['_desk'];
+
+[
+	15, // time in seconds
+	_desk, // arguments for Finish + fail
+	{ // on finish
+		[_this] call AE3_interaction_fnc_desk_open;
+	},
+	{}, // on fail
+	"Opening desk" // progress bar text, can be localized
+	{true}, // condition code checked each frame to continue
+	["isNotSwimming"] // conditions for checking interaction
+] call ace_common_fnc_progressBar;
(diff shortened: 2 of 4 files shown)


y0014984:
I also will review this on tuesday.

addons/interaction/functions/fnc_start_desk_close.sqf
```diff
@@ -0,0 +1,23 @@
...
+params ['_desk'];
+
+[
+	15, // time in seconds
+	_desk, // arguments for Finish + fail
+	{ // on finish
+		[_this] call AE3_interaction_fnc_desk_close;
+	},
+	{}, // on fail
+	"Closing desk" // progress bar text, can be localized
```
y0014984, line 20: The comma is missing. `"Closing Desk",`

addons/interaction/functions/fnc_start_desk_open.sqf
```diff
@@ -0,0 +1,23 @@
...
+params ['_desk'];
+
+[
+	15, // time in seconds
+	_desk, // arguments for Finish + fail
+	{ // on finish
+		[_this] call AE3_interaction_fnc_desk_open;
+	},
+	{}, // on fail
+	"Opening desk" // progress bar text, can be localized
```
y0014984, line 20: The comma is missing. `"Opening Desk",`

addons/interaction/functions/fnc_start_desk_open.sqf
```diff
@@ -0,0 +1,23 @@
+/**
+ * Starts process of opening a desk
+ *
+ * Arguments:
+ * 0: Desk <OBJECT>
+ *
+ * Return Value:
+ * None
```
y0014984, line 8: The function should return a bool, like the function you wrapped around. We already use a wrapper that needs the return value.

addons/interaction/functions/fnc_start_desk_close.sqf
```diff
@@ -0,0 +1,23 @@
+/**
+ * Starts process of closing a desk
+ *
+ * Arguments:
+ * 0: Desk <OBJECT>
+ *
+ * Return Value:
+ * None
```
y0014984, line 8: The function should return a bool, like the function you wrapped around. We already use a wrapper that needs the return value.

addons/interaction/functions/fnc_start_desk_close.sqf
```diff
@@ -0,0 +1,23 @@
...
+*/
+
+params ['_desk'];
+
+[
+	15, // time in seconds
+	_desk, // arguments for Finish + fail
+	{ // on finish
+		[_this] call AE3_interaction_fnc_desk_close;
```
y0014984, line 17: `_this` is not correct, as it contains more then the given arguments. You should change this to `_this select 0`, I think. Or take a look into several other implementations of `progressBar`, for example here: https://github.com/y0014984/Advanced-Equipment/blob/dev/addons/power/functions/fnc_turnOffGeneratorAction.sqf

y0014984 (changes requested):
You really should test your code. See my annotations related to your issues.

Because we are already wrapping this function I think that an additional function/wrapper is too much for this minor feature. Please integrate the progress bar directly into `desk_open` and `desk_close`.

Also keep in mind, that the returned bool is processed to decide if the status of the desk (open/close) has changed. So the return value should depend on wether the progress bar runs in success (completed successfully --> `true`) or get's canceled (via ESC key --> `false`).

Also we should think of the use case that the function could be triggered without a progress bar. To achieve that, please implement a `_silent` mode like in my referenced function.

Do you implement these changes?

y0014984:
I close this PR due to inactivity.

PR #320, opened 2023-01-27
getOpts Syntax Check

y0014984:
Merging this PR will add a Syntax check to every command. This check analyses the syntax definitions of every command and calculates a minimum and a maximum amount of "things" that could/should appear after command itself and the options. In case of a mismatch a message is shown indicating that the player should have a look at the help page by executing command -h

Only a few commands need to be adjusted. This is already done in this PR.


GermanHydrogen (approved):
Looks good

PR #323, opened 2023-01-31
added missing localization strings and updated translations

y0014984:
Merging this PR will add additional localization strings that where missing. Now they can be translated.
Also the russian translations were updated, thanks to Bilibagga.
Also the chinese translations were updated, thanks to @PowerBOXx 
New french translations thanks to alexisdu7589.


y0014984:
Hi @PowerBOXx we have additional texts that need translation. Would you mind to translate them? It's not only the texts in this PR. Would be great if you could look into all stringtable.xml files. We had to change some texts and therefore resetted the translations. Thank you very much.

GermanHydrogen (approved):

PR #327, opened 2023-02-03
Latest Bugfixes v0.5.2

y0014984:
See linked issues for details.

### DIFF
--- a/addons/armaos/functions/fnc_os_echo.sqf
+++ b/addons/armaos/functions/fnc_os_echo.sqf
@@ -34,7 +34,24 @@ private _text = _ae3OptsThings joinString " ";
 
 if (_backslashInterpretion) then
 {
-    _text = _text splitString "\n";
+	_result = [];
+
+	while { true } do
+	{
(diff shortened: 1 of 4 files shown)


addons/armaos/functions/fnc_os_echo.sqf
```diff
@@ -34,7 +34,24 @@ private _text = _ae3OptsThings joinString " ";
 
 if (_backslashInterpretion) then
 {
-    _text = _text splitString "\n";
```
GermanHydrogen, line 37: See https://community.bistudio.com/wiki/BIS_fnc_splitString using the `byWord` option

GermanHydrogen (changes requested):

addons/armaos/functions/fnc_os_echo.sqf
y0014984: Thanks for the suggestion. I've searched for something like that but obviosly you had more luck. I changed this.

GermanHydrogen (approved):

PR #331, opened 2023-02-10
Fix for mv and cp syntax

GermanHydrogen:
Fixed syntax for copying or moving into a folder which already exists, so the trailing slash is not longer necessary


y0014984 (approved):
Looks good to me. You can merge this PR.

PR #336, opened 2023-02-14
Italian keyboard layout and localization

y0014984:
User Seraf asked for italian keyboard layout. He will also help us with the italian localization.

I also added some minor changes to the german keyboard layout and added a missing french translation. These changes would be too small for it's own pull request.


GermanHydrogen (approved):
LGTM

PR #337, opened 2023-02-19
More generators

GermanHydrogen:
Implemented all other vanilla generator objects:
- Radar Generator
- Diesel Ground Power Unit
- Power Generator
- Portable Generator


addons/power/CfgVehicles.hpp
```diff
@@ -154,14 +120,458 @@ class CfgVehicles
...
+		class AE3_Equipment
+		{
+			displayName = "$STR_AE3_Power_Config_PortableGeneratorDisplayName";
+
+			class AE3_ace3Interactions
+			{
+				class AE3_aceDragging
+				{
+					// Dragging
+					ae3_dragging_canDrag = 0;  // Can be dragged (0-no, 1-yes)
```
y0014984, line 523: This generator, like the rugged generator, should be dragable.

addons/power/CfgVehicles.hpp
```diff
@@ -154,14 +120,458 @@ class CfgVehicles
...
+					ae3_dragging_dragPosition[] = {0, 1, 0};  // Offset of the model from the body while dragging (same as attachTo)
+					ae3_dragging_dragDirection = 0;  // Model direction while dragging (same as setDir after attachTo)
+				};
+				class AE3_aceCargo
+				{
+					ae3_cargo_canLoad = 0;  // Enables the object to be loaded (1-yes, 0-no)
```
y0014984, line 529: This generator, like the rugged generator, should be loadable.

y0014984 (changes requested):
The small red generator should have the same features as the rugged example, see comments.

Also the reported output watts are in very high numbers. We should output kilo watts (kW) instead of watts (W) in case of watts above 1.000. That would be earier to read instead of counting zeros.

Also we should start documenting the assets in the wiki. Could you start and create a page containing the features of the generators? I will add all other assets later. Perhaps our initial google spreadsheet will help you on that: https://docs.google.com/spreadsheets/d/1uKQK8no6wsR7n2-Xa8Vi-e6YC-xwoeOwcBT4-IKHopQ/edit?usp=share_link

All after all the PR looks good to me. You can merge the pull request after you made the mentioned minor changes.

y0014984 (approved):
Looks good to me. You can merge this pull request.

PR #340, opened 2023-03-10
UI on texture

y0014984:
### Introduction
Merging this pull request will introduce the **UI on texture** feature, which means that the armaOS UI is rendered also on the laptop texture and can be seen by players in a 3m circle around the laptop. 

There are two triggers that force an update on the **UI on texture**. First, every change on the UI, for example text input/output, changing design, changing battery status etc. forces an update on this specific content. Secondly, an event handler updates all aspects of the **UI on texture** every 5 seconds. This should prevent the case, that someone who joins the circle later, does not get the current content of the UI, if the player that uses armaOS does not interact for a while.

A CBA settings controls wether or not a player invokes updates on his surrounding players textures. Can be forced by server.

### Todos
- [x] add CBA setting to enable/disable this feature globally
- [x] add function headers
- [x] BI Bug Report for the know bug
- [ ] ~add support for `snake`~

### Annotations
- All Updates for UI on texture are send via `remoteExec` to all clients in a 3m circle around the laptop. Even the update for the invoking player is done via `remoteExec`. If this is harmless, I wouldn't change that.
- We could modify the code, so that the `updateAll` function does not fire permanently but only if someone is joining the circle. For now I leave it as it is.
- The new local variable `AE3_UiOnTexActive` is used to determine, if a computer-display was already initialized for "UI on texture" on a specific client. 

### Known Bugs
- The first time a client gets an Update, produces an error. I don't know if this is an Arma 3 Bug or something that I could fix, see screenshot. The Bug is reported here: https://feedback.bistudio.com/T171035

### DIFF
--- a/addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
+++ b/addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
@@ -30,10 +30,13 @@ _computer setVariable ["AE3_terminal", _terminal];
 
 /* ------------- UI on Texture ------------ */
 
-private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
-
+if (AE3_UiOnTexture) then
 {
-   [_computer, _terminalKeyboardLayout] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setKeyboardLayout", _x];
-} forEach _playersInRange;
+	private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+	{
+	[_computer, _terminalKeyboardLayout] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setKeyboardLayout", _x];
+	} forEach _playersInRange;
+};
 
 /* ---------------------------------------- */
--- a/addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
+++ b/addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
@@ -27,3 +27,13 @@ else
 };
 
 _computer setVariable ["AE3_terminal", _terminal];
+
+/* ------------- UI on Texture ------------ */
+
+private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+{
+   [_computer, _terminalKeyboardLayout] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setKeyboardLayout", _x];
+} forEach _playersInRange;
+
+/* ---------------------------------------- */
--- a/addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
+++ b/addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
@@ -34,9 +34,7 @@ if (AE3_UiOnTexture) then
 {
 	private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
 
-	{
-	[_computer, _terminalKeyboardLayout] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setKeyboardLayout", _x];
-	} forEach _playersInRange;
+	[_computer, _terminalKeyboardLayout] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setKeyboardLayout", _playersInRange];
 };
 
 /* ---------------------------------------- */
--- a/addons/armaos/functions/fnc_terminal_setTerminalDesign.sqf
+++ b/addons/armaos/functions/fnc_terminal_setTerminalDesign.sqf
@@ -45,12 +45,15 @@ ctrlSetFocus _consoleOutput;
 
 /* ------------- UI on Texture ------------ */
 
-private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
-
-private _computer = _consoleOutput getVariable "AE3_computer";
-
+if (AE3_UiOnTexture) then
 {
-   [_computer, _bgColorHeader, _bgColorConsole, _fontColorHeader, _fontColorConsole] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setTerminalDesign", _x];
-} forEach _playersInRange;
+   private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+   private _computer = _consoleOutput getVariable "AE3_computer";
+
+   {
+      [_computer, _bgColorHeader, _bgColorConsole, _fontColorHeader, _fontColorConsole] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setTerminalDesign", _x];
+   } forEach _playersInRange;
+};
 
 /* ---------------------------------------- */
--- a/addons/armaos/functions/fnc_terminal_setTerminalDesign.sqf
+++ b/addons/armaos/functions/fnc_terminal_setTerminalDesign.sqf
@@ -42,3 +42,15 @@ private _consoleOutput = _consoleDialog displayCtrl 1100;
 
 // set focus to text field, otherwise focus stays on button and prohibits additional text input
 ctrlSetFocus _consoleOutput;
+
+/* ------------- UI on Texture ------------ */
+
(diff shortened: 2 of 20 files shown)


y0014984:
Possible Workaround to my error is to display the UI the normal way on surrounding clients first. With that the images are loaded and do not produce additional errors if shown with the UI2texture feature. Perhaps I can show and hide the UI fast enough, so the player won't see it at all. Or perhaps there is a hide feature, so the UI is technically displayed but not shown to the player. I will try to implement this workaround.

addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
```diff
@@ -26,4 +26,17 @@ else
...
\ No newline at end of file
+_computer setVariable ["AE3_terminal", _terminal];
+
+/* ------------- UI on Texture ------------ */
+
+if (AE3_UiOnTexture) then
+{
+	private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+	{
```
GermanHydrogen, line 37: You can give the broadcast targets directly to remoteExec:
```
[_computer, _terminalKeyboardLayout] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setKeyboardLayout", _playersInRange];
```

addons/armaos/functions/fnc_terminal_setTerminalDesign.sqf
```diff
@@ -41,4 +41,19 @@ _consoleCtrl ctrlSetTextColor _fontColorConsole;
...
+
+if (AE3_UiOnTexture) then
+{
+   private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+   private _computer = _consoleOutput getVariable "AE3_computer";
+
+   {
+      [_computer, _bgColorHeader, _bgColorConsole, _fontColorHeader, _fontColorConsole] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_setTerminalDesign", _x];
+   } forEach _playersInRange;
```
GermanHydrogen, line 56: See above

addons/armaos/functions/fnc_terminal_uiOnTex_addUpdateAllEventHandler.sqf
```diff
@@ -0,0 +1,54 @@
...
+                private _fontColorConsole = ctrlTextColor _consoleCtrl;
+
+                private _terminal = _computer getVariable "AE3_terminal";
+
+                private _terminalBufferVisable = _terminal get "AE3_terminalBufferVisable";
+                private _size = _terminal get "AE3_terminalSize";
+
+                {
+                    [_computer, _terminalBufferVisable, _size, _terminalKeyboardLayout, _bgColorHeader, _bgColorConsole, _fontColorHeader, _fontColorConsole, _value] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_updateAll", _x];
+                } forEach _playersInRange;
```
GermanHydrogen, line 47: See above

addons/armaos/functions/fnc_terminal_uiOnTex_addUpdateAllEventHandler.sqf
```diff
@@ -0,0 +1,54 @@
...
+ * 1: Computer <OBJECT>
+ * 2: Console <DIALOG>
+ *
+ * Results:
+ * None
+ */
+
+params ["_computer", "_consoleDialog"];
+
+private _updateInterval = 5;
```
GermanHydrogen, line 14: Why do we need to update regularly? Isn't everything pushed when it changes?

addons/armaos/functions/fnc_terminal_updateBatteryStatus.sqf
```diff
@@ -48,10 +48,23 @@ _handle =
...
+
+                /* ------------- UI on Texture ------------ */
+
+                if (AE3_UiOnTexture) then
+                {
+                    private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+                    {
+                        [_computer, _value] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_updateBatteryStatus", _x];
+                    } forEach _playersInRange;
```
GermanHydrogen, line 60: See above

addons/armaos/functions/fnc_terminal_updateBatteryStatus.sqf
```diff
@@ -48,10 +48,23 @@ _handle =
                 };
 
                 _batteryCtrl ctrlSetText format ["\z\ae3\addons\armaos\images\AE3_battery_%1_percent.paa", _value];
+
+                /* ------------- UI on Texture ------------ */
+
+                if (AE3_UiOnTexture) then
```
GermanHydrogen, line 54: Maybe only push an update, when the symbol changed

addons/armaos/functions/fnc_terminal_updateOutput.sqf
```diff
@@ -29,4 +29,17 @@ private _output = [];
...
+
+/* ------------- UI on Texture ------------ */
+
+if (AE3_UiOnTexture) then
+{
+	private _playersInRange = [3, _computer] call AE3_main_fnc_getPlayersInRange;
+
+	{
+		[_computer, _output] remoteExec ["AE3_armaos_fnc_terminal_uiOnTex_updateOutput", _x];
+	} forEach _playersInRange;
```
GermanHydrogen, line 42: See above

addons/armaos/functions/fnc_terminal_uiOnTex_addUpdateAllEventHandler.sqf
```diff
@@ -0,0 +1,54 @@
...
+
+_handle = 
+    [
+        {
+            (_this select 0) params ["_computer", "_consoleDialog"];
+
+            if (AE3_UiOnTexture) then
```
GermanHydrogen, line 21: Why apply the per frame handler, if the feature is disabled?

GermanHydrogen (changes requested):
First of all: Nice work, awesome feature.

I think the way in which you are using the cba setting variable AE3_UiOnTexture, 
does only mean, that a client with uiOnTexture turned off does not broadcast the updates to other players. In case of updateOutput the update push depends on the server setting.
A way to solve this, would be to include a filter to getPlayersInRange instead, to only get players which have the feature enabled. This would also reduce traffic. 
The other possibility, would be to check directly in the uiOnTex update functions.


I did not test the feature in dedicated multiplayer.

addons/main/functions/fnc_getPlayersInRange.sqf
```diff
@@ -0,0 +1,21 @@
...
+ * 
+ * Return:
+ * Nothing
+ */
+
+params ["_range", "_object"];
+
+private _allPlayers = [] call BIS_fnc_listPlayers; // Only players, not headless clients
+
+private _playersInRange = [];
```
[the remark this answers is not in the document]
y0014984: Will use your suggestion.

addons/armaos/functions/fnc_terminal_setKeyboardLayout.sqf
y0014984: Sounds good. Will do that.

addons/armaos/functions/fnc_terminal_uiOnTex_addUpdateAllEventHandler.sqf
y0014984: Yes, everything is pushed, but if player a uses the Terminal and anly reads it without any interaction, there is nothing to push. If player b enters the circle, he won't see the actual content/interface. Imagine a player is not sitting in front of his real life computer and blocking the laptop, or thinking very long about the next command or perhaps unconciousness. That's why there will be an auto-push-interval.

addons/armaos/functions/fnc_terminal_updateBatteryStatus.sqf
y0014984: Yes, you are right. Will try to implement that.

GermanHydrogen (approved):
LGTM

PR #341, opened 2023-03-14
Eden attribute for PowerOn on mission start

GermanHydrogen:
Added an eden attribute to power devices to power them on, on mission start.


y0014984:
The previously described bug appeared in singleplayer. In hosted multiplayer it worked and also the lamp was on.

Despite that we should be able to guarantee the order of execution.

y0014984:
If the lamp is created in Editor before the generator, the lamp will be turned on before the generator is turned on. At this moment there is no power source and the lamp stays off.

A3-Root:
Fixed in PR #426

PR #343, opened 2023-03-31
File Encryption in addFile Module

y0014984:


addons/filesystem/stringtable.xml
```diff
@@ -121,6 +121,60 @@
                 <French>Si le contenu du fichier est une fonction ou un code exécutable depuis le terminal</French>
                 <Italian>Se il contenuto del file è una funzione o codice che può essere eseguito dal terminale</Italian>
             </Key>
+            <Key ID="STR_AE3_Filesystem_Config_IsEncryptedDisplayName">
```
GermanHydrogen, line 124: "Is encrypted?" suggests, that the user input is already encrypted. I think the usage of just the verb "encrypt?" would
be better

GermanHydrogen (changes requested):
The attribute section of the addFile module becomes a bit crowded.
It would be nice, if we could split that up, but I dont know if this is possible

addons/filesystem/stringtable.xml
y0014984: Yeah, you are right. But I don't really like questionmarks in UI. So, I think I will try the imperative form: encrypt content (Or something like that)

y0014984:
New dependency: This PR needs to be changed to reflect changes made in this PR: https://github.com/y0014984/Advanced-Equipment/pull/344

This adds an additional encryption method.

addons/filesystem/functions/fnc_moduleAddFile.sqf
```diff
@@ -32,13 +33,59 @@ private _permissions = [
...
 
+if (_isEncrypted) then 
+{
+	private _mode = "encrypt";
+
+	_content = _content splitString endl;
+
+	private _row = "";
+
+	{
```
GermanHydrogen, line 55: Why do you changed the encryption loop. I don't understand, why you now check and inititalise the encryption method for each row seperatly

y0014984: The "columnar" algorithm also encrypts the line breaks, meaning line breaks on wrong positions. That leads to disturbed text can't be decrypted because the length of the message needs to match the password and you don't know where a line would end. So partially decrypting isn't possible. That beeing said my approach was to not encrypt the complete text but line by line. Therefore I need a loop somewhere. If the loop isn't outside of the switch statement, I would need to loops, one in each case statement. As we say in germany: "Nicht schön, aber selten." Any suggestions to improve to code are welcome.

GermanHydrogen: As we say in germany: "Vorschlag zur Güte" ^^ (Not tested)
```
private _crypto_fnc;
switch (_encryptionAlgorithm) do
{
	case "caesar":
	{
		_encryptionKey = _encryptionKey call BIS_fnc_parseNumber; // needs a number
		_encryptionKey = round _encryptionKey; // needs an integer
		_crypto_fnc = AE3_armaos_fnc_encryption_caesar;
	};
	case "columnar":
	{
		_crypto_fnc = {
			params ["_encryptionKey", "_mode", "_row"];
			_row = _row regexReplace [" ", "_"];

			while {(count _encryptionKey) < 2 } do 
			{
				// min. length 2
				_encryptionKey = _encryptionKey + "_";
			};
			_crypto_fnc = AE3_armaos_fnc_encryption_columnar;
		}
		
	};
};

{
	_row = [_encryptionKey, _mode, _x] call _crypto_fnc;
	_content set [_forEachIndex, _row];
} forEach _content;
```

y0014984:
"Land_Portable_generator_F" --> "$STR_A3_cfgVehicles_Land_Portable_generator_F0"
"Land_PowerGenerator_F" --> "$STR_A3_CfgVehicles_Land_PowerGenerator_F0"
"Land_MobileRadar_01_generator_F" ---> "$STR_A3_C_CfgVehicles_Land_MobileRadar_01_generator_F0"
"Land_PortableGenerator_01_sand_F" --> "$STR_A3_C_CfgVehicles_Land_PortableGenerator_01_sand_F0"
"Land_PortableGenerator_01_black_F" --> "$STR_A3_C_CfgVehicles_Land_PortableGenerator_01_black_F0"
"Land_PortableGenerator_01_F" --> "$STR_A3_CfgVehicles_Land_PortableGenerator_01_F0"

addons/filesystem/functions/fnc_moduleAddFile.sqf
y0014984: Looks elegant. I've adapted your suggestion. Thank you. Please review the changes.

GermanHydrogen (approved):
LGTM

PR #344, opened 2023-04-06
Add columnar transposition cipher

JulesVerner:
This PR adds the columnar transposition cipher as an alternative for the already implemented caesar cipher.
So far, it's only an additional algorithm-option for the `crypto`-command and does _not_ have an implementation for the `crack`-command.
I added a corresponding wiki-suggestion in the wiki folder; Although the explanation might be a bit complicated.
Looking forward to hearing your thoughts on this! 😄


y0014984:
Thank you for this contribution. I will have a look into it tomorrow.

y0014984:
I forgot to mention: please add all languages to the stringtable.xml. You can use the english version but please add the tags for all currently supported languages. This is easier for our translators und prevents formatting issues. Thank you.

JulesVerner:
> * Use blank spaces to fill rows on encryption and decryption instead of random chars

I would suggest using '_' or similar special characters, instead of spaces.
Spaces appended to the end of the message are not visible in the console and therefore it's likely that a false encrypted message will be used.

y0014984:
> I would suggest using '_' or similar special characters, instead of spaces.

Do it. We will see how it behaves.

y0014984 (approved):
Looks good to me. I will merge this PR.

PR #399, opened 2023-11-16
Updated hemtt version to 1.10.0

GermanHydrogen:


addons/main/CfgUserInterfaceZeus.hpp
```diff
@@ -949,7 +949,7 @@ class AE3_UserInterface_Zeus_Module_AddFile
 
             checked = 0;
 
-            onCheckedChanged = "params ['_control', '_checked']; private _display = ctrlParent _control; private _algorithmCtrl = _display getVariable ['algorithmCtrl', objNull]; private _keyCtrl = _display getVariable ['keyCtrl', objNull]; if (_checked == 1) then { _checked = true; } else { _checked = false; }; _algorithmCtrl ctrlEnable _checked; _keyCtrl ctrlEnable _checked;";
+            onCheckedChanged = "params ['_control', '_checked']; private _display = ctrlParent _control; private _algorithmCtrl = _display getVariable ['algorithmCtrl', objNull]; private _keyCtrl = _display getVariable ['keyCtrl', objNull]; if (_checked == 1) then { _checked = 'true'; } else { _checked = false; }; _algorithmCtrl ctrlEnable _checked; _keyCtrl ctrlEnable _checked;";
```
y0014984, line 952: I think this should stay `true` instead of `'true'` because it's SQF.

addons/main/CfgUserInterfaceZeus.hpp
```diff
@@ -1501,7 +1501,7 @@ class AE3_UserInterface_Zeus_Module_AddConnection
 
             text = "$STR_AE3_Main_Zeus_Switch";
 
-            onButtonClick = "params ['_control']; private _display = ctrlParent _control; private _switch = _display getVariable ['switch', false]; if (_switch) then { _switch = false; } else { _switch = true; }; _display setVariable ['switch', _switch]; private _fromCtrl = _display displayCtrl 1401; private _toCtrl = _display displayCtrl 1402; private _fromText = ctrlText _fromCtrl; private _toText = ctrlText _toCtrl; _fromCtrl ctrlSetText _toText; _toCtrl ctrlSetText _fromText;";
+            onButtonClick = "params ['_control']; private _display = ctrlParent _control; private _switch = _display getVariable ['switch', false]; if (_switch) then { _switch = false; } else { _switch = 'true'; }; _display setVariable ['switch', _switch]; private _fromCtrl = _display displayCtrl 1401; private _toCtrl = _display displayCtrl 1402; private _fromText = ctrlText _fromCtrl; private _toText = ctrlText _toCtrl; _fromCtrl ctrlSetText _toText; _toCtrl ctrlSetText _fromText;";
```
y0014984, line 1504: The same as above. I think this should stay `true` instead of `'true'` because it's SQF.

.hemtt/project.toml
```diff
@@ -0,0 +1,31 @@
+name = "Advanced Equipment"
+prefix = "ae3"
+mainprefix = "z"
+
+[properties]
```
y0014984, line 5: Are these properties necessary? author seems to be hardcoded everywhere and template ... I don't know. Looks kind of optional to me if this isn't used as a variable that is referenced in some other places I think. How ACE3 does it? They changed to HEMTT v1.0 recently.

.hemtt/project.toml
```diff
@@ -0,0 +1,31 @@
...
+template = "cba"
+
+#folder_optionals = true
+#sig_version = 3
+
+[lint.sqf]
+enabled = true
+
+[version]
+git_hash = 0
```
y0014984, line 16: I thought that this could be ignored if you use `script_version.hpp` instead for versioning.

.hemtt/project.toml
```diff
@@ -0,0 +1,31 @@
+name = "Advanced Equipment"
+prefix = "ae3"
+mainprefix = "z"
```
y0014984, line 3: I can't find `mainprefix` in the HEMTT documentation. If I understand the documentation correctly `prefix` could be used as a variable in the `$PBOPREFIX$` files. Not sure.

y0014984 (changes requested):
Build worked. Mostly it looks good to me. See my annotations. Please change the 2 occurances where you put `true` in quotation markes but it's SQF. All other annotations are optional.

.hemtt/project.toml
y0014984: ACE3 does this the same way.

.hemtt/project.toml
y0014984: ACE3 does this the same way.

.hemtt/project.toml
y0014984: ACE3 does this the same way.

addons/main/CfgUserInterfaceZeus.hpp
GermanHydrogen: Fixed

addons/main/CfgUserInterfaceZeus.hpp
GermanHydrogen: Fixed

y0014984 (approved):
Looks good to me.

PR #402, opened 2023-12-19
changed keyboard input to multilingual

### LINKED ISSUE #135: additional keyboard layouts
1. british, spanish, french, italian
2. polish, czech, serbic
3. russian, chinese, japanese, corean

### LINKED ISSUE #235: Change Terminal Control fromt RscText to RscEdit?
Perhaps this could give us a lot of options and opportunites back, because this would simulate a cursor, wie can make selections and we don't need separate Keyboard Layouts for every country. This would work, if we could capture every keypress in the RscText field and to stop it's effects if unwanted, for example the up- and down-keys, because the user shouldn't be able to leave the current input line. I need to check that.

### LINKED ISSUE #241: Add new Keyboard Layouts
We should add more keyboard layouts so people around the world could use our mod or we should try to switch to RscText, see #235 so there is no need for individual keyboard layouts. This would save a lot of time and would reduce errors.

### LINKED ISSUE #326: Add CAPSLOCK support to armaOS
At the moment all keyboard input in treated directly ignoring if CAPSLOCK is currently active. We should enable that feature.

### LINKED ISSUE #335: Character Length > 1 for some symbols in armaOS terminal
I encountered a problem while creating new keyboard layouts for armaOS terminal. Some symbols, that could be entered like the "€" symbol, need more then one backspace keypress to delete them. Some symbols need 2 and some 3 keypresses. 

This looks like that the encoding of the keys is UTF-8, which in some cases is represented by 1 8-bit character, and in some cases by 1 16-bit character, especially if the character is not part of the ASCII alphabet table. For example the german "ü" sign needs two backspace presses, the "€" sign needs 3 backspace presses.

Perhaps we can somehow work around this problem if SQF provides text functions to work around that behavior.

y0014984:
Merging this pull request will change the keyboard input. It's now handled via an invisible input text field that processes the input independently fromt the used keyboard layout. Then I copy the resulting characters to our terminal buffer. Now there is no need anymore for the corresponding CBA setting and switching keyboard layouts. All code belonging to keyboard switching was removed.

I still need to check how this will behave for eastern users like chinese and how do they input via the IME


addons/armaos/functions/fnc_terminal_updateOutput.sqf
```diff
@@ -26,8 +26,10 @@ private _output = [];
 	_output pushBack _buffer;
 } forEach _terminalBufferVisible;
 
-_outputControl ctrlSetStructuredText (composeText _output);
-ctrlSetFocus _outputControl;
+_consoleOutput ctrlSetStructuredText (composeText _output);
+
+private _consoleInput = _consoleOutput getVariable ["AE3_consoleInput", controlNull];
+ctrlSetFocus _consoleInput;
```
GermanHydrogen, line 32: Why do we need to set the focus again?
If this is important, please at a comment

GermanHydrogen (changes requested):
A very elegant solution for the keyboard layout problem. Overall LGTM, besides one minor comment. 
In the future we should refactor `fnc_terminal_addEventHandler` since it is quite long and convoluted IMHO.

addons/armaos/functions/fnc_terminal_updateOutput.sqf
y0014984: @GermanHydrogen  The focus changes in cases where the player clicks on other UI elements like the buttons in the header. I don't think that it's strictly necessary but this way we ensure that the focus is on the input field.

A3-Root:
This has been abandoned.

PR #403, opened 2023-12-20
added 'tree' command

y0014984:
Merging this Pull Request will add the 'tree' command to armaOS.

### DIFF
--- a/addons/armaos/functions/fnc_os_tree.sqf
+++ b/addons/armaos/functions/fnc_os_tree.sqf
@@ -1,0 +1,103 @@
+/**
+ * Prints the filesystem tree with the current working directory as root to stadout.
+ *
+ * Arguments:
+ * 1: Computer <OBJECT>
+ * 2: Options <[STRING]>
+ * 3: Command Name <STRING>
+ *
+ * Results:
+ * None
+ */
+
+params ["_computer", "_options", "_commandName"];
+
+private _commandOpts = [];
+private _commandSyntax =
+[
+	[
+			["command", _commandName, true, false]
+	]
+];
+private _commandSettings = [_commandName, _commandOpts, _commandSyntax];
+
+[] params ([_computer, _options, _commandSettings] call AE3_armaos_fnc_shell_getOpts);
+
+if (!_ae3OptsSuccess) exitWith {};
+
+private _pointer = _computer getVariable "AE3_filepointer";
+private _filesystem = _computer getVariable "AE3_filesystem";
+
+private _current = [_pointer, _filesystem] call AE3_filesystem_fnc_resolvePntr;
+
+private _terminal = _computer getVariable "AE3_terminal";
+private _user = _terminal get "AE3_terminalLoginUser";
+
+private _level = 0;
+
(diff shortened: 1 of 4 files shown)


addons/armaos/functions/fnc_os_tree.sqf
```diff
@@ -0,0 +1,103 @@
...
+    try
+    {
+        // function does not return a bool but throws an exception if permissions insufficient
+        [_filesystem, _user, _permissionNeeded] call AE3_filesystem_fnc_hasPermission;
+
+        {
+            // Hashmap forEach variables: KEY = _x && VALUE = _y
+
+            private _treeString = "";
+            for "_i" from 0 to _level do
```
GermanHydrogen, line 59: Could be simplified to

```sqf

for "_i" from 0 to _level - 1 do
{
    _treeString = _treeString + "|  ";
};
_treeString = _treeString + "|--";
```

addons/armaos/functions/fnc_os_tree.sqf
```diff
@@ -0,0 +1,103 @@
...
+private _filesystem = _computer getVariable "AE3_filesystem";
+
+private _current = [_pointer, _filesystem] call AE3_filesystem_fnc_resolvePntr;
+
+private _terminal = _computer getVariable "AE3_terminal";
+private _user = _terminal get "AE3_terminalLoginUser";
+
+private _level = 0;
+
+private _fnc_tree_recursive =
```
GermanHydrogen, line 38: Subfunction should be defined at the top of the function for consistency

addons/armaos/functions/fnc_os_tree.sqf
```diff
@@ -0,0 +1,103 @@
...
+};
+
+private _result = [_pointer, _current, _user, _level] call _fnc_tree_recursive;
+
+private _totalResults = _result select 0;
+private _missingPermissions = _result select 1;
+
+if (_missingPermissions > 0) then
+{
+	_totalResults append [format [localize "STR_AE3_ArmaOS_Exception_CantScanFolderMissionPermissions", _missingPermissions]];
```
GermanHydrogen, line 98: Alternatively, we could simply hide the contents of dirs for which the user does not have permission. I don't know what would be more consistent.

GermanHydrogen (changes requested):
Looks good in general, only small comments

addons/armaos/functions/fnc_os_tree.sqf
y0014984: Looks good. Will implement that.

addons/armaos/functions/fnc_os_tree.sqf
y0014984: Ok. Will do that.

addons/armaos/functions/fnc_os_tree.sqf
y0014984: I copied the behavior from the `find` command. There we do the same. So I think this is consistent.

GermanHydrogen (approved):
LGTM

PR #404, opened 2023-12-20
added new font 'UbuntuMono'

y0014984:
Merging this Pull Request will add a custom font to armaOS. I created the font files with the FontToTGA-tool from the Arma 3 Tools. I modified the cfg file to create additional glyphs that are necessary for the 'tree' command. 

class Custom
{
  start = 0x2500;
  end = 0x257F;
};

The font is from https://fonts.google.com/specimen/Ubuntu+Mono and I used this manual to create the font files https://forums.bohemia.net/forums/topic/179963-tutorial-import-font-on-arma-3/

A custom font will allow us to use the correct glyphs for the tree command and to create a custom cursor instead of the return symbol we are currently using. But first this PR needs to be merged.


GermanHydrogen (commented):
The changes LGTM. Can you check how much this increases the file size of the final pbo? Would be weird if would double our mod size with a font

y0014984:
Mod sizes:
dev branch             9,56 MB
new-font branch    14,9 MB  (increase of 56 %)

y0014984:
I deleted the sizes 27 to 46 because I think we don't need them. The sizes 6 to 26 seem to be sufficient for our needs. This decreases the mod size to 12,6 MB (increase of 31,8 %)

I can live with that. What about you, @GermanHydrogen ?

GermanHydrogen (approved):
LGTM

PR #406, opened 2023-12-21
changed interaction to only one interaction point

y0014984:


addons/flashdrive/functions/fnc_initInterface.sqf
```diff
@@ -11,72 +11,83 @@
...
-	private _actions = [];
-	private _class = configFile >> "CfgWeapons" >> "Item_FlashDisk_AE3";
-
-	private _add = {
-		params ["_target", "_player", "_params"];
-		([_target, _player] + _params) remoteExec ["AE3_flashdrive_fnc_connectFlashDrive", 2];
+if(!isDedicated) then
+{
+	// get or add parent action
+	private _parentActionPath = _device getVariable ["AE3_parentActionPath", []];
```
GermanHydrogen, line 17: Code duplication with addParentAction

addons/flashdrive/functions/fnc_initInterface.sqf
```diff
@@ -11,72 +11,83 @@
...
+	if (_parentActionPath isEqualTo []) then
+	{
+		private _parentAction = ["AE3_ParentAction", _name, "", {}, {true}] call ace_interact_menu_fnc_createAction;
```
GermanHydrogen, line 20: Where is `_name` defined?

addons/interaction/functions/fnc_initLaptop.sqf
```diff
@@ -22,15 +22,13 @@ else
 
 if(!isDedicated) then
 {
-    // add the armaOS Menu to ACE3 Interaction Menu
-    private _armaOSAction = ["AE3_ArmaOSAction", localize "STR_AE3_ArmaOS_Config_ArmaOSDisplayName", "", {}, {true}] call ace_interact_menu_fnc_createAction;
-    [_laptop, 0, ["ACE_MainActions"], _armaOSAction] call ace_interact_menu_fnc_addActionToObject;
+    private _parentActionPath = _equipment getVariable ["AE3_parentActionPath", []];
```
GermanHydrogen, line 25: Why don't u use addParentAction instead?

addons/network/functions/fnc_initNetworkDevice.sqf
```diff
@@ -35,36 +35,38 @@ private _childs =
...
+					{
+						params ["_target", "_player", "_params"]; 
+						_params params ["_device"];
+						(alive _target) and (!isNull (_device getVariable ["AE3_network_parent", objNull]))
+					},
+					{},
+					[_entity]
+					] call ace_interact_menu_fnc_createAction;
+	
+	private _parentActionPath = _entity getVariable ["AE3_parentActionPath", ""];
```
GermanHydrogen, line 66: Why don't u use addParentAction instead?

addons/network/functions/fnc_initRouter.sqf
```diff
@@ -37,53 +37,54 @@ private _childs =
...
+					{
+						params ["_target", "_player", "_params"]; 
+						_params params ["_device"];
+						(alive _target) and (!isNull (_device getVariable ["AE3_network_parent", objNull]))
+					},
+					{},
+					[_entity]
+					] call ace_interact_menu_fnc_createAction;
+
+	private _parentActionPath = _entity getVariable ["AE3_parentActionPath", ""];
```
GermanHydrogen, line 68: Why don't u use addParentAction instead?

addons/power/functions/fnc_initBattery.sqf
```diff
@@ -18,48 +18,50 @@ private _entity = _battery;
...
+					_params params ["_battery"];
 					_handle = [_battery] spawn AE3_power_fnc_checkBatteryLevelAction;
 				}, 
 				{alive _target},
 				{},
 				[_battery]
 				] call ace_interact_menu_fnc_createAction;
 
-	[_entity, 0, ["ACE_MainActions", "AE3_DeviceAction"], _check] call ace_interact_menu_fnc_addActionToObject;
+	private _parentActionPath = _entity getVariable ["AE3_parentActionPath", ""];
```
GermanHydrogen, line 38: Why don't u use addParentAction instead?

addons/power/functions/fnc_initGenerator.sqf
```diff
@@ -24,8 +24,10 @@ if(!isDedicated) then
 				{params ["_target", "_player", "_params"]; _handle = [_target] spawn AE3_power_fnc_checkPowerOutputAction;}, 
 				{alive _target}] call ace_interact_menu_fnc_createAction;
 
-	[_entity, 0, ["ACE_MainActions", "AE3_DeviceAction"], _check] call ace_interact_menu_fnc_addActionToObject;
-	[_entity, 0, ["ACE_MainActions", "AE3_DeviceAction"], _power] call ace_interact_menu_fnc_addActionToObject;
+	private _parentActionPath = _entity getVariable ["AE3_parentActionPath", ""];
```
GermanHydrogen, line 27: Why don't u use addParentAction instead?

addons/power/functions/fnc_initPowerInterface.sqf
```diff
@@ -81,8 +59,32 @@ if(_internal) then
...
+					{
+						params ["_target", "_player", "_params"]; 
+						_params params ["_device"];
+						(alive _target) and (!isNil {_device getVariable "AE3_power_powerCableDevice"})
+					},
+					{},
+					[_generator]
+					] call ace_interact_menu_fnc_createAction;
+
+	private _parentActionPath = _device getVariable ["AE3_parentActionPath", ""];
```
GermanHydrogen, line 84: Why don't u use addParentAction instead?

addons/power/functions/fnc_initSolarPanel.sqf
```diff
@@ -11,21 +11,23 @@
...
 
 if(!isDedicated) then
 {
 	private _power = ["AE3_PowerAction", localize "STR_AE3_Power_Interaction_CheckPowerOutput", "", 
-				{params ['_target', '_player', '_params']; _handle = [_target] spawn AE3_power_fnc_checkPowerOutputAction;}, 
+				{params ["_target", "_player", "_params"]; _handle = [_target] spawn AE3_power_fnc_checkPowerOutputAction;}, 
 				{alive _target}] call ace_interact_menu_fnc_createAction;
 
-	[_entity, 0, ["ACE_MainActions", "AE3_DeviceAction"], _power] call ace_interact_menu_fnc_addActionToObject;
+	private _parentActionPath = _entity getVariable ["AE3_parentActionPath", ""];
```
GermanHydrogen, line 22: Why don't u use addParentAction instead?

GermanHydrogen (changes requested):
Pulling all the action functions inside the !isDedicated conditionals worsens the readability of the overall functions. The performance price for that should be negligible.

addons/power/functions/fnc_initPowerInterface.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/power/functions/fnc_initSolarPanel.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/power/functions/fnc_initGenerator.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/power/functions/fnc_initBattery.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/network/functions/fnc_initRouter.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/network/functions/fnc_initNetworkDevice.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/interaction/functions/fnc_initLaptop.sqf
y0014984: I don't have access to the `_name` variable. To achieve that I would need to pass additional parameters to this function. Because of the way the code flows, this variable is already set and can easily be used.

addons/flashdrive/functions/fnc_initInterface.sqf
y0014984: That's true. The USB interfaces are a special case. I think I need to pass additional variables to treat it like in the other cases. Also, for the future, we should rewrite our config-compile-init-system. It's quite a mess with duplicated logic in 3 different addons.

addons/flashdrive/functions/fnc_initInterface.sqf
y0014984: Nowhere. It's missing. You are right. I need to rewrite some parts.

y0014984:
See my changes for review.

Now that all interactions are in one place we can decide what is better: Option A (all together) or Option B (grouped by Addon). How do you think about it @GermanHydrogen 

AE3-Interactions-Decision

addons/power/functions/fnc_initBattery.sqf
GermanHydrogen, line 38: Is the default a valid input for `ace_interact_menu_fnc_addActionToObject`?

We should decide if we want to recover here if  `AE3_parentActionPath` is not set, so to just use the "ACE_MainActions" as a parent, or if we want to throw an exception in this case

GermanHydrogen (commented):
The one commend applies for all power device init functions

addons/flashdrive/functions/fnc_initInterface.sqf
```diff
@@ -4,79 +4,84 @@
...
 		{
-			private _action = [_name, _name, "", {}, _condition, _flash_drives, [_y]] call ace_interact_menu_fnc_createAction;  
-			_actions pushBack [_action, [], _target];
-		};
-	} forEach _interfaces;
-
-	_actions;
-};
+			params ["_target", "_player", "_params"];
+			(_params select 0) params ["_index", "_name", "_rel_pos", "_rot_yaw", "_rot_pitch", "_rot_roll"];
```
mrschick, line 54:
```suggestion
			private _driveInfo = _params select 0;
			if !(_driveInfo isEqualType []) then {
				_driveInfo = _params select 1;
			};
			_driveInfo params ["_index", "_name", "_rel_pos", "_rot_yaw", "_rot_pitch", "_rot_roll"];
```
This change fixes #414, details [here](https://github.com/y0014984/Advanced-Equipment/issues/414#issuecomment-2052427187).

mrschick (changes requested):

addons/flashdrive/functions/fnc_initInterface.sqf
mrschick: Thanks to Dystopian's https://github.com/acemod/ACE3/pull/9946, this rough fix won't be necessary, the next ACE3 Patch version will resolve the issue.

mrschick: This has now been fixed publicly by [ACE v3.17.1](https://github.com/acemod/ACE3/releases/tag/v3.17.1).

PR #426, opened 2026-07-25
Public RELEASE v2.0.0.0

### LINKED ISSUE #9: armaOS chat functionality

### LINKED ISSUE #10: armaOS network file transfer
If we have implemented the armaOS network, we could use it to transfer files from on computer to another.

### LINKED ISSUE #17: checking on/off-state and battery level like ACE3 for ammunition
If you check the ammunition with ACE, there isn't something like 12/30 bullets left, but a magazine icon with a fill bar, so you do not know the exacly value. This could be adapted to check the capacity of external batteries, laptops and the generator and similar items. I would say that knowing the exact value should only be possible if the device has a complex interface, like if you are logged into the laptop. But if the laptop is closed, there should be only a rudimentary indicator.

### LINKED ISSUE #44: Filesystem Fill Module
For example user folders for all opfor AI names in a 250 m radius. Or typical system folders like system, bin, sbin, temp, local, users, applications, data, etc.

### LINKED ISSUE #45: Show Filesystem Objects that are path to image
distinguish between text content and image path by path divider / or \

### LINKED ISSUE #51: armaOS: Not so strict case sensitivity

### LINKED ISSUE #61: Network Management

### LINKED ISSUE #62: login/shh to another computer in this network

### LINKED ISSUE #64: armaOS: New Command hostname (shows local or remote hostname)

### LINKED ISSUE #72: armaOS: ping 127.0.0.1 should work as expected

### LINKED ISSUE #73: armaOS vi or any simple text editor

### LINKED ISSUE #76: Testing: What happens on canceling ACE progess bars

### LINKED ISSUE #78: ArmaOS: Chat while loop
https://github.com/y0014984/Advanced-Equipment/blob/bff1f3f93c06970bd0a10eb0d19ecbb60172dd9d/addons/main/functions/armaos/fn_chat.sqf#L68

Syntax Error: Round brackets instead of curly brackets.

### LINKED ISSUE #104: Allow putting small items like laptop, battery pack or usb-stick into backpacks
If the items are small enough, it should be possible to carry them around, not only with ACE3 carrying but also as an inventory item in your backpack (not uniforms or vests). 

A good example for this are the ACE3/ACEX intel items.

### LINKED ISSUE #105: Visual helper for wired connections
If you connect devices you need a visual hint of how long is the "invisible" cable. This could be achieved with an local indicator, that shows the max. cable length, like some translucent sphere around the object or something like that.

### LINKED ISSUE #106: Add turn on/off sound to lamps
Esspecially these 2 assets (Land_PortableLight_single_F_AE3 and Land_PortableLight_double_F_AE3) could use a sound, indicating the turn on and turn off action. Especially if the generator turns off or too much devices pull power, a sound indicating the loss of power at the lamp would be nice. I already searched for some adequate sounds on freesound.org. Now I need to modify them and implement it into the mod.

### LINKED ISSUE #107: Power output indicator for power generator
We already have an option to check the remaining fuel in the power generator. But we also need an option to check the power output to prevent pulling to much power and with that turning of the generator and all connected devices. So an option that says "Power Output 4370/5000 Watts" would be nice.

### LINKED ISSUE #108: E-Mail instead of chat for ArmaOS
Perhaps an E-Mail-Client would be easier to implement instead of a chat client. A virtual e-mail-server could run on every client. Messages could be addressed by using unsername@machine 

For example if user alex uses laptop milsim1 and user tom uses computer milsim2 then alex could write a message to tom:

FROM: <alex@milsim1>
TO: <tom@milsim2>
MESSAGE: Your next objective is Kavala. Advance.

Alex could write this e-mail with an specific client. The message is forwarded to tom if milsim2 is reachable. If not, the message is stored and alex needs to retry later.

If the message is delivered to milsim1, tom get's a message after login or in his active terminal session, that there is a new message. This is a file in a specific postbox folder in the userfolder that tom can access. He can read the message with cat or some text editor.

Adding e-mail-files is an interesting opportunity for mission creators, so a hacker could investigate a laptop/computer and finds interesting information in these e-mails, like names, hostnames, timestamps, content etc.

### LINKED ISSUE #109: timestamps for ArmaOS filesystem
We should implement timestamps for filesystem objects. This becomes very useful for hacking research on other laptops/computers. Especially if timestamps on e-mail-files are investigated, so you have some kind of temporal order.

These timestamps should be printed with the long version of ls

### LINKED ISSUE #110: Use initial values for lamp animations (extent, pitch, yaw)
The 3 Lamp Assets with extractable lamps have specific start values for all 3 modification types (extend, pitch, yaw), but these start values are ignored in our implementation. They all start with 0 or something like that. So we should read these values initially and store them so they can be used in the animation as the start value.

### LINKED ISSUE #113: Restrict movement area while adjusting lamps/solar panels
If you adjust the lamps or solar panels you are free to move around. You need some freedom of movement but you should not be able to move very far away from the asset while still adjusting. So we need to check that and restrict the movement to a certain area or stop the adjustment interface if you are too far away from the asset.

### LINKED ISSUE #114: Seperate Animations for both sides of the desk
For now if you unwrap the desk, both table sides show up. Perhaps it is more convenient to allow to unwrap both sides of the desk independantly. Should be easy to implement.

### LINKED ISSUE #116: ACE3 Progress Bar for unfolding/folding rugged desk

### LINKED ISSUE #124: Cleanup deprecated code
There is some deprecated code, that isn't nessesary anymore after our rewrite of most parts. I will try to find and cleanup these parts because I know best whats old and whats not.

### LINKED ISSUE #125: Should we move all assets into a separate asset addon and consolidate config parsing?
Our asset config definitions are widely spread over our addons. Laptop is in armaos, generator and battery pack is in power framework and lamps etc. are in interaction addon. Perhaps it's cleaner to move all assets in a seperate asset addon. The asset config parsing mechanism can still be part of the seperate addons but should be merged together. right now there are two separate config parsers. They are build in a way that they are not mergeable. But I think we could join them together if the individual parser checks if there is already a partially pared config, so he will only add additional config informations instead of creating new ones.

### LINKED ISSUE #126: Asset texture animations for rugged assets
Multiple of our used assets have some kind of interface. Some have blinking lights, some have displays showing the current ingame time oder the power level. We should find a way to manipulate and unse there interface elements. Perhaps we create complete new assets to achive this. Or we device to ignore this feature because the implementation is too time consuming.

### LINKED ISSUE #129: ArmaOS Multipage-Output vs. Scrolling Feature
We already have a scrolling feature so the user can see all output even if the output of a ArmaOS command is longer than the terminal output size. But there could be another mechanism that provides long outputs page by page. I personally don't like page-wise output but the could be valid reasons for that. We need to discuss this.

### LINKED ISSUE #132: ArmaOS blinking cursor
If we have an individual terminal font with an full size block, we could implement a blinking cursor.

### LINKED ISSUE #134: Unique Hostnames
The default Hostname for armaOS Laptop is "armaOS". We should set unique hostnames on mission init. This could be relevant later, for the e-mail feature or some network commands.

### LINKED ISSUE #135: additional keyboard layouts

### LINKED ISSUE #137: Change user in an active Terminal session
It could be interesting to be able to change the active user in a running terminal session, perhaps with a chuser command. For now you need to logout and login again. The pain isn't that big so this issue has a low priority.

### LINKED ISSUE #141: sudo command
To execute commands as root, we should implement the sudo command, because logging in as root will be disabled. Therefore we need a setting to define, which users are in the sudoers group. Also we need the command itself.

### LINKED ISSUE #146: Typing Animation for characters using a laptop/armaOS
I think there is a typing animation that we could use for a player character while he is using the laptop/armaOS/computer.

### LINKED ISSUE #148: ArmaOs Terminal: Copy/Paste
Implement copy and paste shortcuts in armaos terminal.

https://community.bistudio.com/wiki/copyToClipboard
https://community.bistudio.com/wiki/copyFromClipboard

### LINKED ISSUE #156: Power network load
While most of the power calculations is done server side, most of the 
variables are broadcasted to all clients every calculation cycle. This could put unnessary strain
on the server/network if a lot of clients are connected to the server.

Could be inproved by only updating the necessary variables on the server and implementing getter
functions, in the case the clients need to check something.

### LINKED ISSUE #169: In-Game Manual
Like in other mods it would be helpful if the player could find some basic information on how to use Advanced Equipment Objects. Players are often confronted with new mods and new functionality and whilst in-game they don't know how to use it. The map menu would be a good place to add some information about AE3.

### LINKED ISSUE #170: armaOS diagnostics command for objects
This is a very complex one. The idea is that some information about devices is only available if you connect the device with a computer and then access the device's internal data via a "diagnostics" command in armaOS. Perhaps this would also allow to manipulate settings or trigger functions of the device, that would be otherwise hidden. Perhaps this could be useful if you need a security password to access the device.

### LINKED ISSUE #175: change "is binary" file property to "filetype", to distinguish between: text, image, executeable

### LINKED ISSUE #176: add "is Admin" option to addUser module

### LINKED ISSUE #180: allow linking the content of files added by addFile module to files in mission folder

### LINKED ISSUE #193: add searching for substrings and file content to 'find' command
At the moment find searches only for full/complete file/folder names. We could allow using `*` to search for substrings. Also we could allow to search for file contents with an additional command parameter like `-c` for content search.

### LINKED ISSUE #195: Snake game im armaOS terminal window with characters
Coding Example: https://www.youtube.com/watch?v=AaGK-fj-BAM

### LINKED ISSUE #196: AE3 transistor radio
Idea for an additional device. Battery powered. Plays songs from sound files configured via module or in device settings. Possible settings/interactions: turn on/off, volume, track selection, shuffle, radio mode (songs configured event wide via module, that a received with every radio in the mission --> could be used by mission makers for breaking news), built-in batteries, charge via power providing devices or no built-in batteries.

### LINKED ISSUE #197: Interaction Icons for ACE interactions
- computer
- power
- network

### LINKED ISSUE #198: Hint to indicate successful connection between devices
An additional hint would be helpful to indicate that a new power or network connection between devices was successfully established. So the user does not need to lookup the connection status immediately after creating the connection.

### LINKED ISSUE #206: add Computer without battery
check dependencies like the battery symbol.

### LINKED ISSUE #209: change 'ipconfig' to 'ifconfig' or 'ip' command
As one user mentioned, `ipconfig` is a windows command. We are simulating a linux, so we should change that to `ifconfig` or `ip -a` / `ip -r`

Originally I tried to create a unique armaOS, that mixes commands from different operating systems but nowerdays it seems more convenient to simply simulate a typical and simplified linux experience.

### LINKED ISSUE #212: armaOS header after login
Perhaps we should change when the header appears in armaOS. Compared to something like Ubuntu, the header appears after successful login, not bevore that.

### LINKED ISSUE #213: english teaser/trailer video

### LINKED ISSUE #214: add a change password command and/or a 'change your password' feature
If the mission maker creates a mission and adds users to a laptop, he knows about the passwords and they need to be somehow disclosed to the user, so he can use it. 

It could be a good idea to allow the user to change his own user passwor. With that he can keep the password secret, even the mission maker doesn't knows about. 

An alternative or additional feature could be, that the user needs to change his password at the first login. So he probably remembers his own password better.

I a worst case the mission maker or zeus should be able to recover a password, perhaps via a special command or a zeus module.

### LINKED ISSUE #215: boxloader support/compatibility
A user suggested to use or support the [Boxloader Mod](https://steamcommunity.com/sharedfiles/filedetails/?id=1199318917). This would be an interesting feature as it combines the logistics ideas. But for now I think it's a little bit out of scope and would need a lot of development resources. For now we are using ACE3 Cargo system. Perhaps in the far future we could think about and discuss this topic. I only want to keep track of that idea.

### LINKED ISSUE #219: Add Name Attribute to AE3 Eden Editor Attributes
We already have fuel and battery level (unused at the moment). We could also add an name attribute. This name could be set via init of the asset to set the ACE3 Cargo name. With this feature a mission maker could pre-define the names of devices, to keep track the easy way. Otherwise you need to name them on ever mission startup.

### LINKED ISSUE #220: Power & Lan network visualization in debug overlay
Adding a network visualization in the debug overlay via drawLine3D.

### LINKED ISSUE #230: Terminal line breaks and Cyrillic Glyphs
See screenshots. It looks like our automatic line breaks in the armaOS Terminal break the lines way too early if cyrillic glyphs are used. Perhaps this is related to Unicode. We need to investigate further.

### LINKED ISSUE #231: Input line breaks and disappearing cursor
On verly long input lines that break to the next line, there is one step where the cursor disappears.

### LINKED ISSUE #232: Use animation states from Eden Editor as initial states
Some of our equipment, like the desk or the lamps, allow in Eden Editor to change their animation states. At the moment we ignore what is set up in Eden Editor and set our own animation states. But we could read and use them, so what you see in Eden Editor and what you see in mission is in sync.

### LINKED ISSUE #235: Change Terminal Control fromt RscText to RscEdit?

### LINKED ISSUE #236: Input of accent glyphs
Some glyphs need to be created by pressing two different keys after each other. For example, to print the character `é` you first press the  key ``` which results in no printed sign, secondly you press the `e` button and then these tho signes are merged together. At the moment the input of such signs is not possible in our Terminal because we can only print characters that need one keypress.

We could implement this by adding some logic for this glyphs or we switch to a normal `RscEdit` control, like mentioned in issue #235

### LINKED ISSUE #237: Translation of texts in mod.cpp
Do translations work there? It's outside of a pbo folder.

### LINKED ISSUE #238: Add special AE3 event handlers
For mission makers it could be useful to use special AE3 event handlers to trigger their own stuff. Possible Event Handlers could be:

- Fuel empty/full
- Device on/off
- Connection between devices created/removed
-

### LINKED ISSUE #242: Add new Crypo Algorythms
To enhance the crypto/cracking gameplay there should be more then one crypto algorythm. The is a good overview of some simple algorythms at http://www.crypto-it.net/eng/simple/index.html 

3-5 algorythms should be sufficient.

### LINKED ISSUE #245: Fix Debug Overlay
Because getPowerOutput and getBatteryLevel use the getRemoteVar Function, the Server is called onEveryFrame twice for each AE3 power asset. This needs to be fixed bevore Debug Overlay could be used in an dedicated server environment. 

The draw needs to be on each frame, but could use a local variable to display it's content. This local variable could be updated less often, like once per second. This would reduce server load by factor of 20 or more. 

Also we need to find a way to use getBatteryLevel and getPowerOutput in a unscheduled environment.

### LINKED ISSUE #248: Add ACE3 Interaction status and network connections to debug overlay
- dragActive
- carryActive
- cargoActive
- inUse
- turnedOn
- unwrapped
- powerConnected
- networkConnected
- networkConnectionsCount

### LINKED ISSUE #250: Check in Eden Editor for allowed Module Connections
Perhaps some users will connect the armaOS Modules (addDir, addFile or addUser) not directly to a computer but perhapts to other modules or unsupported assets. So we could perhaps check in Eden Editor for correct connections. The are [Eden Editor Edent Handlers](https://community.bistudio.com/wiki/Arma_3:_Eden_Editor_Event_Handlers), that perhaps could be useful in this case, like: `OnConnectingStart` and `OnConnectingEnd` or `ConnectionChanged3DEN`

We only need to figure out how to prohibit or remove the connection.

### LINKED ISSUE #257: Switch rpt log entries to the CBA/ACE style
There must be a mechanism that creates the typical RPT log file entries. Here is an example with mod name, addon name and debug level. 

```
18:19:26 [ACE] (common) INFO: Extension version: ace_advanced_ballistics: 3.12.0-8ddde18
18:19:26 [CBA] (xeh) INFO: [877,34.646,0] PostInit finished.
18:19:26 [CBA] (optics) INFO: Scripted camera restarted.
18:19:26 [ACE] (common) INFO: Settings initialized.
18:19:26 [ACE] (common) INFO: 1 delayed functions running.
```

These are perhaps the CBA macros LOG(), INFO(), WARNING() and ERROR() but I need to figure out how to use them. Probably they use a global DEBUG Level, see here: https://cbateam.github.io/CBA_A3/docs/files/main/script_macros_common-hpp.html#DEBUG_MODE_FULL

### LINKED ISSUE #263: ACE3 rename is disabled while ACE3 cargo is deactivated
This is not exactly an issue. It's more the or less the way it works. Perhaps we could make a Pull Request for ACE3 that allows renaming while ACE3 cargo is disabled.

While this is not changed, we should mention this as a "known issue".

### LINKED ISSUE #266: Non-reachable catch for spawn execution and obsolete _result in 'executeFile'
https://github.com/y0014984/Advanced-Equipment/blob/0bec49ca28c7d177e08c84dc74f75cd57774c90f/addons/armaos/functions/fnc_shell_executeFile.sqf

In this function all commands of armaOS are spawned. Because of `spawn` the catch clause will never be reached via a thrown exception from within a command. Either we only use try/catch for calling `getFile` or we find a way to process exceptions at this point that are thrown by a spawned function.

Additionally, the defined `_result` variable is not used anywhere and the calling function `fnc_shell_process.sqf` does not process any return values. So this line can be removed.

### LINKED ISSUE #267: 'DEL' key to delete in Terminal while working with the moveable cursor
@GermanHydrogen , perhaps that's easy for you to add, because you created the moveable cursor. At the moment we can delete text in the armaOS Terminal with direction to the left (BACKSPACE key) but we can't use the DEL key to delete to the right. Prior to the 'movable cursor' feature there was no need to delete to the right but now it would be an additional nice feature.

### LINKED ISSUE #268: Adjusting sitting position for rugged chair
It's already possible to sit on the rugged chair but the sitting position is not perfect. We should adjust the position so it doesn't look like the players as is hovering above the chair.

### LINKED ISSUE #269: A user reported problems using AE3 and Alive mod
It's a comment on Steam Workshop. The user said:

25. Nov. um 20:26 
whenever I try to go into a mission made with ALiVE, it breaks and doesn't spawn any units, removed this mod and then everything works perfectly fine

25. Nov. um 20:24 
hey man, idk if you knew, but your mod is breaking ALiVE. Could you please check it out?

### LINKED ISSUE #270: Thunder, Storms, Rain and Electricity
We could, perhaps, as a feature that there is a chance that a lightning strikes into a AE3 power device, perhaps a generator under bad weather conditions like storm and rain.

### LINKED ISSUE #271: Fingerprint sensor for authentication on armaOS
Instead of using a username and a password we could allow using a fingerprint sensor. This should be easy to implement. We could add a module for allowing fingerprint sensor logins. Syncing module to computer, syncing players to module. If module is present, the login is skipped and the playername becomes the username including a home directory. This is has the advantage that users don't need to remember logins for their own computers.

### LINKED ISSUE #272: random function for creating random textbased mini-games
If players write their own armaOS function they perhaps want some mini puzzles to use out of the box to create something that needs to be solved bevore getting any further. The level should be adjustable. A single call inside a custom command should be the goal.

### LINKED ISSUE #274: Fold/Unfold action for flexible solar panel
We now use this thing only in the unfolded thate but there is an additional model for the folded state. We could also allow putting the flexible solar panel into players inventory like a backpack.

### LINKED ISSUE #275: Debug Overlay adjustments
We need to remove the hard coded classes and add additional information. Also we could draw lines between connected devices.

### LINKED ISSUE #276: Desk wings block player
The foldable desk wings block player from moving at their position, even if their are folded in. I think that's a restriction we can't modify but we should keep track of that.

### LINKED ISSUE #277: Rename or rewrite turn on/off/standby functions
They now indicate to be usable with every computer but they are closely related to the first laptop. They should be usable more dynamically.

### LINKED ISSUE #278: Set ACE3 Cargo name while in Eden Editor

### LINKED ISSUE #279: Add new interactions to see the amount or list of connected devices
At the moment, if you connect for example multiple lamps to the generator, you can't keep track of which devices are connected to the generator. You need to check one by one. An easier way would be to allow checking the amount of connections and perhaps the names of connected devices via an interaction at the generator.
Also an additonal action would be useful to disconnect all currently connected devices at once. This relates to all devices, power or network, that establish 1:N connections, like generator, solar panel, battery pack, router and so on.

### LINKED ISSUE #280: 'caesar' encryption: allow additional signs and symbols
At the moment only the 26 upper case characters of the latin alphabet are allowed for the `caesar` algorythm, used in the `crack` and `crypto` commands, but additional characters would be useful.

- `0-9` would allow to easily encrypt informations like coordinates
- `-` and `_` would allow to use other word delimiters, so it becomes harder to decode

Adding these additional characters we would end up with 36 characters: 
`ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_`
With more allowed characters the bruteforce attack takes longer, but 36 is not a problem. But if more players want even more allowed symbols, perhaps also lowercase, then this becomes a problem if you want a useable bruteforce attack.  

Or we could do something different: Keep the 26 characters but add the 10 numbers via a parallel encryption, so you can see in the encrypted text, that at this position is a encrypted number. So, if you have a key of 3, an `X` encrypts to a `A` and a `9` encrypts to a `2`. This would keep the simplicity of the algorythm, but also allows encryption of numbers.

We also could allow every other sign and keep it as it is, so it's not encrypted, like we do at the moment with the ` ` space character. 

I need to think about that. Any opinions?

### LINKED ISSUE #283: Add blank line before next command prompt line
It's hard to distinguish between multiple similar commands and their outputs, so we should give visual aid. In windows terminal and perhaps in linux terminal too, there is a blank line before the next command prompt. Example:

```
admin@armaOS:/home/admin>ls -al
file1
file2

admin@armaOS:/home/admin>
```

### LINKED ISSUE #284: Processing commands with whitespaces in paths
At the moment all commands and options are splitted via:

```sqf
private _commandSegments = _input splitString " ";
```

So it is not possibile to process paths that contain whitespaces/blanks. If we want to allow filenames and dir-names containing whitespaces, then we should insert an additional processing step that parses for `"` or "'" and then rejoins some command segments like:

(Pseudo code)
```sqf
private _newCommandSegments = [];
private _stringBuffer = "";
private _quotationMarksCounter = 0;

{
   
   if ((((_x select [0, 1]) isEqualTo """) || ((_x select [0, 1]) isEqualTo """)) && (_quotationMarksCounter == 0)) then
   {
      _stringBuffer = _stringBuffer + (_x select [1, (count _x) - 1];
      _quotationMarksCounter = 1;
      continue;
   };

   if (_quotationMarksCounter == 1) then
   {
      _stringBuffer = _stringBuffer + _x;
      _quotationMarksCounter = 1;
      continue;
   };

   if (!(((_x select [(count _x) + 1, 1]) isEqualTo """) || ((_x select [(count _x) + 1, 1]) isEqualTo """)) && (_quotationMarksCounter == 1)) then
   {
      _stringBuffer = _stringBuffer + (_x select [0, (count _x) - 1];
      _quotationMarksCounter = 0;
      _commandSegment pushBack _stringBuffer;
      _stringBuffer = "";
      _continue;
   };

   if (_quotationMarksCounter == 0) then
   {
      _commandSegment pushBack _x;
   };

} forEach _commandSegment;
```

But we could also decide to not allow whitespaces/blanks in the entire armaOS workflows. In this case we should check all posibilities to create files and dirs (config, modules and commands) and replace whitespaces/blanks by an replacement symbol like `_` underscore. Also this needs to be documented.

### LINKED ISSUE #285: Refactoring power functions
A refactoring on power functions would be helpful. Unfortunately, there are for example some variables indicating a generator but instead this var means beeing a power device connected to a generator, for example in `fnx_initPowerInterface.sqf` 

Every time I read that code my brain get's problems. So for future enhancements a refactoring would safe a lot of time understanding what's going on.

### LINKED ISSUE #286: Code redundancy enhancement for ACE3 settings
In PR #243 I introduce a new variable called `AE3_SettingsACE3`

The content of this variable is somewhat redundant because we already saved these settings in a config var while parsing the config. So perhaps this could be simplified.

### LINKED ISSUE #288: Add a way to change the address space of routers
At the moment rounters are initialized with 192.168.0.1 but there is no way to change that. Via script the ip address of the router could be changed but we should also think of mission makers that want an easier way. This is very important if you want to connect multiple routers. In this case the address spaces should be different. Possible solutions are:

### Current state via script
```sqf
private _address = [192, 168, 0, 1];
_router setVariable ["AE3_network_address", _address, true];
```

### Alternative solutions

- Set IP address via Eden Attribute, like with fuel level and power level
- Set IP address via additional Module (set ip address module)
- Set IP address via function call in init field `[this, [192, 168, 0, 1]] call AE3_network_fnc_setIpAddress;`
- ...

### LINKED ISSUE #289: Add command 'grep'
A user asked to add the `grep` command to armaOS. With reduced abilities I think we can implement that because SQF supports regular expressions.

### LINKED ISSUE #299: Generator is enterable by using ACE3
@GermanHydrogen said, that it's still possible to enter the generator "vehicle" by using ACE3. We need to investigate and fix that.

### LINKED ISSUE #301: isCyclic check for custom 3DEN connections (network and power)
We should check, if the created custom connections in 3DEN for power and network are cyclic. Cyclic network connections are already eliminated on mission init but it would be useful for mission creators to see already in the 3DEN Editor that a connection is cyclic. In that case the connection should be removed.

### LINKED ISSUE #302: Add Network Config like for Power Config
At the moment network devices are not defined via AE3 config. So filtering for these classes is quite difficult and at some points even hard coded like in the connection checks for the custon 3DEN Editor connections. We should change to config that is parseable so we can remove the hard coded content.

### LINKED ISSUE #312: Change all single quotes to double quotes where not needed
We should only use single quotes if necessary, for example for nested quotes. In normal use cases we should use only double quotes.

### LINKED ISSUE #314: Generator spawns with crew in Eden Editor
Because our Generator is a Vehicle, in the Eden Editor the default check mark for spawning with crew is enabled. Therefore the generator spawns with crew and has a blue marking indicating that it belongs to blufor. If you disable the checkmark, the generator has a yellow marking indicating an empty vehicle. 
I don't know at the moment if this could lead to some unexpected behavior, like enemies attacking the generator. Perhaps we should try to change/fix that nevertheless.

### LINKED ISSUE #316: Add new grafic functions and games to retro environment
## Functions
- [ ] setBackgroundColor
- [ ] drawLine
- [ ] drawRect
- [ ] drawCircle

## Games
- [ ] moon patrol
- [ ] space invaders
- [ ] pong (vs. AI?)
- [ ] mine sweeper
- [ ] pac man
- [ ] tetris

### LINKED ISSUE #326: Add CAPSLOCK support to armaOS

### LINKED ISSUE #328: Allow closing the Laptop in Eden Editor
At the moment the open/close state is loaded for each laptop from config and will be set accordingly on mission startup. We could allow to set the state in the Eden Editor attributes for that object. I think this is not hard to implement. Also we could update the model in the Eden Editor if the state is set/changed.

### LINKED ISSUE #332: USB-Stick is not available in ACE3 Arsenal
The Flash Drive seems to be available in Zeus but not in ACE3 Arsenal.

### LINKED ISSUE #335: Character Length > 1 for some symbols in armaOS terminal

### LINKED ISSUE #352: Allow definition of filesystem objects in mission config
Adding files and directories via modules is sufficient for a few filesystem objects. But if you want to build a big filesystem with a lot of objects, that is very laborious. Additionally, handling filesystem permissions could be a problem with the automatically generated parent folders. Because the processing order of the modules is not fixed, this could lead to issues that behave differently every time you start your mission. You could write scripts to add filesystem objects but not everybody likes that. 

So, we think it would be a good idea to allow definition of filesystem objects via mission config. We already parse config for filesystem objects for our default objects. I think we could adapt that mechanism.

### LINKED ISSUE #354: Change Asset Config Definitions for Color variants
Our asset configs are very long especially because of the color variants. We should try to use morge base classes. This would reduce the overall amount of config code. See the generators config in #350 as an example.

### LINKED ISSUE #357: New addons for assets, zeus and eden editor
To organize things better, we could introduce new addons and move related code to them. This could be assets, zeus, eden editor and more.

### LINKED ISSUE #359: Introduce new variable to check if an asset hat a specific feature
Currently I'm struggling always with the task to decide wether or not a given entity is for example a computer or not. Yes, we want to keep things simple and flexible, but I write a lot of code to check for things like that. Perhaps we could introduce some variables like:

- hasTerminal = isComputer = hasArmaOS
- hasFilesystem
- isNetworkClient
- hasUSB
- hasBattery
- hasFuelTank

and so on.

### LINKED ISSUE #360: Change internal permissions order from dxrwxrw to drwxrwx
drwxrwx is the default standard and used in our UI, but the mixedup version dxrwxrw is used in our code. I always need to lookup which one is correct. We should simplify and harmonize that.

### LINKED ISSUE #361: Add Eden Editor Module Checks for correct type of synched objects
Currently you could sync the addFile, addDir, addUser etc. Modules to every asset. That doesn't make sense. We should add checks for that.

### LINKED ISSUE #362: Add information about outgoing/incoming power/network connections to Zeus UI
We can add connections but then the visualization disappears. Either I am able to create custom connections or we should be able to lookup existing connections.

### LINKED ISSUE #365: IP address reset for clients if router is turned off
We should reset the IP address of every network consumer (client) to `[127, 0, 0, 1]` if the router is turned off. Currently there is no turn off function defined in config. We need to distinguish between regular clients and other routers. We should prevent resetting the IP address of a router.

Currently the only reset will be done if the network connection is removed. Otherwise the IP address stays the same, even if the router is turned off.

### LINKED ISSUE #367: Add mechanism to determine for all assets if they are initialized by all modules
Zeus functions use this workaround to see, if the asset is initialized completely. This does not work for desk, because this is no power asset. We need a mechanism that checks if an asset is initialized by all modules, no matter which modules are responsible for the current asset.

As a workaround we currently set this variable in the `initDesk` function.

```sqf
waitUntil { !isNil { _entity getVariable "AE3_power_hasInternal" }; };
```

### LINKED ISSUE #372: change addConnection zeus module to work like the ACE3 Load in Cargo module (custom connection)

### LINKED ISSUE #373: improve addConnection module icon (Zeus)

### LINKED ISSUE #375: add new Zeus module to remove and show existing power/network connections

### LINKED ISSUE #378: Automatic update for sys file content in /sys/battery/capacity
The content of the file should show the current battery capacity. I forgot to update this within the `AE3_armaos_fnc_terminal_updateBatteryStatus` function. Should be easy to implement.

### LINKED ISSUE #384: Tunr off error message in dedicated MP
TurnOff on "Diesel Ground Power Unit" with error message in hosted MP, not in Singleplayer oder dedicated MP; Only if router is connected to generator, not if laptop is connected; Somehow related to turnOffWrapper, updateSelfPower, UpdatePower, getPowerOutput; Perhaps server-side error, which can only be seen in hosted MP (because host == server) which is related to devices, that will be turned off because the generator also turns off; reproduceable; error can be found in server-log

### LINKED ISSUE #387: Add error sound
When someone uses a armaOS command wrong or if an script error occurs we could play an error sound, perhaps like the windows error sound.

### LINKED ISSUE #395: Check KeyDown Event Handler Behavior
At some point of testing with the Gruppe Adler modset I experienced a behavior, when pressing the 'u' key while typing in the armaOS terminal, the Gruppe Adler Group Management interface popped up. Currently I can't reproduce that behavior. But perhaps we need to investigate that. 

A possible solution could be to create a display instead on an dialog. I think this way the displayEventHandler of the default display that hosts the dialog would be ignored. Currently the text control of the terminal window uses a controlEventHandler instead of an displayEventHandler.

### LINKED ISSUE #397: Add File Size and Copy Speed
To simulate the transfer/copying of very large file we could simulate a file transfer speed that depends on file size. Perhaps we should itroduce a dummy/blob file type that has size and an overall setting for lokal / USB / Network transfer speeds.

### LINKED ISSUE #401: Adding a Crash State to armaOS
At the moment we can only normally turn on/off the computer. How about adding a new state for crashes so the computer becomes unuseable. 

While implementing this we could repair the "silent" turn off functions.

### LINKED ISSUE #405: Check code for places that need 'forceUnicode'
Unfortunately, string length counts bytes but not characters. That means that unicode characters count 2 bytes instead of one. In scripts you can change this behavior with the [forceUnicode](https://community.bistudio.com/wiki/forceUnicode) command. Affected commands are:

- [copyFromClipboard](https://community.bistudio.com/wiki/copyFromClipboard)
- [copyToClipboard](https://community.bistudio.com/wiki/copyToClipboard)
- [count](https://community.bistudio.com/wiki/count)
- [find](https://community.bistudio.com/wiki/find)
- [in](https://community.bistudio.com/wiki/in)
- [insert](https://community.bistudio.com/wiki/insert)
- [reverse](https://community.bistudio.com/wiki/reverse)
- [select](https://community.bistudio.com/wiki/select)
- [splitString](https://community.bistudio.com/wiki/splitString)
- [trim](https://community.bistudio.com/wiki/trim)
- [regexFind](https://community.bistudio.com/wiki/regexFind)
- [regexMatch](https://community.bistudio.com/wiki/regexMatch)
- [regexReplace](https://community.bistudio.com/wiki/regexReplace)

So, every time we use one of these commands and the affected string could be a user generated unicode character we should take account for that. I have already done that at one place in PR #402 for the function `fnc_terminal_removeCharFromInput.sqf`

### LINKED ISSUE #417: AddFile Module: Prevent encryption of Code-Files
That doesn't make any sense. Shouldn't be possible.

### LINKED ISSUE #423: Filesystem not ready. Please wait and try again.
On dedicated server, with 9 people ingame. While in Zeus placed down rugged laptop, and tried to add user module. Module appeared, stayed there for a few seconds, maybe 10-15 and gave me "Filesystem not ready. Please wait and try again." messege.
Tried all other modules after that and spawning new rugged laptop. Same error.

### LINKED ISSUE #424: UI on texture not working properly
The text doesn't appear, while battery consumption and keyboard layout does.

Running on dedicated server with Headless Client.

[Here's a clip](https://drive.google.com/file/d/1mMAR_7ImvVyRkUk3l-9LbbsMJb4U6lLT/view?usp=sharing)

A3-Root:
Latest GUI update and fixes the following issues:

Closes #424
Closes #423
Closes #417
Closes #405
Closes #401
Closes #397
Closes #395
Closes #387
Closes #384
Closes #378
Closes #375
Closes #373
Closes #372
Closes #367
Closes #365
Closes #362
Closes #361
Closes #360
Closes #359
Closes #357
Closes #354
Closes #352
Closes #335
Closes #332
Closes #328
Closes #326
Closes #316
Closes #314
Closes #312
Closes #302
Closes #301
Closes #299
Closes #289
Closes #288
Closes #286
Closes #285
Closes #284
Closes #283
Closes #280
Closes #279
Closes #278
Closes #277
Closes #276
Closes #275
Closes #274
Closes #237
Closes #272
Closes #271
Closes #270
Closes #269
Closes #268
Closes #267
Closes #266
Closes #263
Closes #257
Closes #250
Closes #248
Closes #245
Closes #242
Closes #238
Closes #237
Closes #236
Closes #235
Closes #232
Closes #231
Closes #230
Closes #220
Closes #219
Closes #215
Closes #214
Closes #213
Closes #212
Closes #209
Closes #206
Closes #198
Closes #197
Closes #196
Closes #195
Closes #193
Closes #180
Closes #176
Closes #175
Closes #170
Closes #169
Closes #156
Closes #148
Closes #146
Closes #141
Closes #137
Closes #135
Closes #134
Closes #132
Closes #129
Closes #126
Closes #125
Closes #124
Closes #116
Closes #114
Closes #113
Closes #110
Closes #109
Closes #108
Closes #107
Closes #106
Closes #105
Closes #104
Closes #78
Closes #76
Closes #73
Closes #72
Closes #64
Closes #62
Closes #61
Closes #51
Closes #45
Closes #44
Closes #17
Closes #10
Closes #9


A3-Root:
Recommended to **SQUAD AND MERGE** rather than merge commit.

y0014984 (approved):
I don't really have the time to review this.

y0014984:
> Recommended to **SQUAD AND MERGE** rather than merge commit.

Can you merge on your own?

